PR #11, opened 2020-05-21
reduce GeoJSON file size

kovaca:
Hi, great boundaries. Love to see it. 

Submitting this PR with a suggested tweak to reduce the size of GeoJSON output files 30-40% without sacrificing any practical degree of precision. 

The GeoJSON format specifications suggest [6ish decimal places as a common default](https://tools.ietf.org/html/rfc7946#section-11.2): (precision of ~10cm). Other sources elaborate on the [practical limits of decimal degree precision](https://gis.stackexchange.com/a/8674).  

Passing [parameters from OGR](https://gdal.org/drivers/vector/geojson.html#layer-creation-options) to the fiona Shapefile -> GeoJSON file creation step achieves this with minimal changes to the code. Currently defaults to 15 decimal places, which is unnecessary. Suggesting 7 here to save some bytes. Hoping this can cut down a smidge on bandwidth and storage needs. 

Cheers!

### DIFF
--- a/buildStage3.py
+++ b/buildStage3.py
@@ -235,10 +235,11 @@ def pBuild(row):
   #even though it's unnecessary for many.
   srcShp.schema["geometry"] = "MultiPolygon"
 
+  kwargs = {"COORDINATE_PRECISION":7}
   with fiona.open(geojson, 'w', driver="GeoJSON", 
                   schema=srcShp.schema,
                  encoding='utf-8',
-                 crs=fiona.crs.from_epsg(4326)) as write_geojson:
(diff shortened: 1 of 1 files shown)


DanRunfola:
Hey kovac - awesome, love to see this, though I pity the soul that is digging through my code :).  I'm pulling this in.

I am just about to do a big '3.0' push, so will incorporate this change in that.  FYI, 3.0 will also include a topology-preserving simplified version of geoBoundaries, which has *massive* file size savings.

kovaca:
Good luck with the push! I imagine setting coordinate precision could further minify simplified version too. Looking forward to 3.0.

PR #124, opened 2020-08-31
Split metadata into 3 files based on each gB release type

slfuhrig:
This was a significant update that divided the original metadata file into separate files based on release type, as well as removing the now obsolete Google ID column.


DanRunfola:
Prep for 4.0, merged!

DanRunfola:
@all-contributors please add @slfuhrig for code

Issue #139, opened 2020-09-09
gbOPEN- ESP ADM 2


slfuhrig:
Our existing ADM 2 is actually ADM 4, so this one needs to be replaced.

carolinamonrivera:
is this still open?

DanRunfola:
Closed by https://github.com/wmgeolab/geoBoundaries/pull/256

PR #142, opened 2020-09-10
#139

MichaelR-geoLab:
New Spain ADM2


MichaelR-geoLab:
double checked and found a bad license; found a new one with an open license I will upload soon

DanRunfola:
Got it, thanks for the update Michael.

PR #158, opened 2020-09-12
deleted humanitarian files for NGA from gB OPEN

slfuhrig:
The files were not deleted from gB OPEN after being moved to gB Humanitarian.


DanRunfola:
Accepting PR (after closely inspecting this and confirming "NGA" did, in fact, refer to Nigeria and not the US Intelligence Agency 👍

PR #168, opened 2020-09-20
Added NRU_ADM2 to gbOpen metadata

sylviashea1:
Only want to push the change in open metadata. 
The text files were from the tutorial. Deleted them from my local repo, but they're still kicking around. 
"updating local repo..." was from when I tried to update my repo via fetching from upstream/master--not sure what's up with that.


DanRunfola:
Hey Sylvia -

You probably just have to push from your local to your master, that would delete the test files.

That said, as we aren't using gbOpenMetaData.csv any longer, could you take a look at the NRU_ADM2.zip file and update the meta.txt file in there with whatever you wanted to?

PR #179, opened 2020-09-24
Updated NRU_ADM2.zip

sylviashea1:
Replaced old NRU zip with new one, now includes meta.txt and license.png. Also deleted test.txt.


sylviashea1:
My original changes to gbOpen metadata are still in the PR. Let me know if I should delete the file on here-not sure how that would affect master.

DanRunfola:
Yep, just delete.  As long as you have a meta.txt in your zip file it'll all work as it should.

sylviashea1:
-Geometry is incomplete, missing middle coverage
-Attribute table incorrect, missing ISO code
-Meta check failed, but meta.txt and license.png were present in zipfile

PR #253, opened 2020-10-06
added ESP_ADM2 file through website

MichaelR-geoLab:
added canonical boundary name (province).

## Checklist:

- [X] I have included a file named "meta.txt", with all required metadata (see an example at https://github.com/wmgeolab/gbRelease/blob/master/sourceData/gbOpen/AGO_ADM1.zip).


DanRunfola:
Just FYI, you have a folder in your zip file (/ESP_ADM2/).  All files are expected to be in the root.  I.e.:
ZIPFILE:
   ESP_ADM2/
      meta.txt

doesn't work

ZIPFILE:
  meta.txt

will work.  I'll try to update the gbBot to make that error more obvious.  I would do a search for a meta.txt, but that will take a bit because I'll have to handle all sorts of possible conditions (i.e., someonne accidentally has two).

DanRunfola:
Followup moved into PR https://github.com/wmgeolab/gbRelease/pull/256

PR #159, opened 2020-09-13
NER_ADM1 Open Zip

helenabuckman:
Updated metadata spreadsheet and added zip to "gbOpen." Renamed file with date because NER_ADM1 already exists and I didn't want to delete.

PR #341, opened 2020-10-10
NER_ADM1

helenabuckman:
This is the updated version of NER_ADM1 (with changes made by Dan). Issue #159

Previous data was moved humanitarian, open license data needed. 

## Checklist:
- [X ] My submission is either a shapefile (with all the relevant files) or a geojson.
- [X] I have included a file named "meta.txt", with all required metadata (see an example at https://bit.ly/2SAiWHk).


slfuhrig (changes requested):
One of the reasons why this PR is failing checks is that you have two files here, so you need to remove the duplicate (one way to do that is to just do a new PR in a new branch with the correct file). The bot also flagged a license issue. You should pass the metadata check if you remove the quotes from the license type and make sure that you have a link to where you found the data.

slfuhrig:
Addressed by newer PR.

PR #344, opened 2020-10-11
NER_ADM2

carolinamonrivera:
gbHumanitarian MRT ADM 2 metadata validation #330

Validated release, source and added license.png.

https://github.com/wmgeolab/gbRelease/issues/329

## Checklist:

- [x ] I have included a file named "meta.txt", with all required metadata (see an example at https://bit.ly/2SAiWHk).


slfuhrig (approved):
I think everything looks good even though all checks haven't passed.

DanRunfola:
The errors here are coming because you've pushed two files into the PR, instead of just one.  Delete the erroneous one and I think this will be ready to go.

DanRunfola (changes requested):
The errors here are coming because you've pushed two files into the PR, instead of just one. Delete the erroneous one and I think this will be ready to go.

slfuhrig (changes requested):
The release type and other notes fields are what's causing you to fail the metadata check.

slfuhrig:
Replaced with #536

PR #354, opened 2020-10-16
Website: In "Visualizing with Python" example, remove dependency on shapely

karimbahgat:
In "Visualizing with Python" example, remove dependency on shapely. 

Also, switch from scandalous javascript style if-statement to pythonic ;) 

I have not checked that it renders correctly, so please double-check.


DanRunfola (approved):

DanRunfola:
:+1: thank you!

DanRunfola:
@all-contributors please add @karimbahgat  for code

PR #552, opened 2020-10-25
NER_ADM1

helenabuckman:
Re-upload since shapefile would not open. All other data is the same.

Previous shapefile was moved to Humanitarian
## Anything Unusual?
Re-submit because the previous shapefile would not open 

## Checklist:

- [X] I have included a file named "meta.txt", with all required metadata (see an example in https://bit.ly/2GtQ01i).


slfuhrig (changes requested):
Almost there!! Just 2 things: remove the quotes from the license type in metadata, and add in ISO_Code (iso.org should have the codes for NER_ADM1). Let me know if you need help with ISO_Code (it's a manual input).

sylviashea1:
Issues fixed in another PR.

PR #563, opened 2020-10-26
Add slider tool

oforiaddaejnr:
The slider tool release 1.0 is ready to be displayed on the website


DanRunfola (changes requested):
Folder structure needs to be named appropriately - are any of these files > 20mb?  If so we need to consider how we're pushing here.  Could also be confusing having multiple ISO codes in the repo.  Can we add a "naturalEarth" prefix to the files themselves?

DanRunfola:
Rejecting this for now until we have a better draft ready.

DanRunfola:
@all-contributors please add @oforiaddaejnr  for code

PR #571, opened 2020-10-27
NPL_ADM1

neelsimpson:
NPL_ADM1 MetaData Validation done 10/27

## Checklist:


sylviashea1 (changes requested):
Hey Neel! A few things that need to be changed:
1. Please add the ISO_Code field and add the ISO codes 
2. I found the province names [here ](https://en.wikipedia.org/wiki/Provinces_of_Nepal). Please replace the province names with the proper ones!

PR #610, opened 2020-11-03
NER_ADM0

neelsimpson:
Updated Metadata fixed on 11/2

## Checklist:


slfuhrig:
Please add the URL to the license png and add a "Name" field with the full name (Niger) in addition to the "ISO_Code" field with the ISO (NER) @neelsimpson. Can we add this license to the acceptable list? @DanRunfola

DanRunfola:
So - "Creative Commons Attribution" is not a valid license; i.e., there is no license called "Creative Commons Attribution".  It looks like, from license.png, this is a OSM Share-Alike license.

(Also, for OSM cases, we want license source 1 to be Open Street Map, and source 2 to be "wherever we found it", mapcruizin in this case).

DanRunfola:
@all-contributors please add @neelsimpson  for data

PR #617, opened 2020-11-03
MDV_ADM0 Metadata Validation #489 - Corrected

kpupkiewicz:
Previous PR had the zip file in the wrong folder. As mentioned with the previous PR, this metadata validation is using an updated shapefile.

## Checklist:


slfuhrig (changes requested):
Hey Kristina! This is a great shapefile, but unfortunately we cannot use it because it originates from GADM (even though that's source 2). I'm going to close this PR and leave the issue open to tackle next semester.

PR #680, opened 2020-11-10
ISR_ADM1

dorianvmiller:
Metadata validation for ISR_ADM1

 ISR_ADM1

 #409 

## Checklist:


sylviashea1 (approved):
Yeehaw, nice job.

DanRunfola:
Ok... so, this is a hard one.  While I'm going to merge this in (as it's better than what we have), we eventually need to return to ISR, as the boundaries we want in our database should be reflective of "what Israel calls Israel" - i.e., inclusive of Gaza and the West Bank.  We'll then back out the contested area pieces later in the process.

But, for now, :+1:!

Issue #19, opened 2020-06-03
MUS ADM1 / ADM0 Disagreement


DanRunfola:
Mauritius has a large disagreement in our source data between the ADM0 and ADM1 boundaries.  For a future release, we should attempt to identify an authoritative source.  We also need to be specific in that we are mapping *land* boundaries, not sea, which may have other ramifications across our dataset.

nmspage:
![Discrepancy_ADM0_ADM1](https://user-images.githubusercontent.com/68017504/88993306-8de81e80-d2b3-11ea-998b-6e694b3ccb1f.PNG)

It appears that ADM 1 is missing two islands to the north of Mauritius. I will extract those features and add them to ADM 1.

PR #695, opened 2020-11-13
MUS_ADM0, MUS_ADM1 #19

nmspage:
 #19

## Checklist:


slfuhrig (changes requested):
SOOO close! Just change the ADM1 ISO_Codes to match the political units (from iso.org)

DanRunfola:
Good robot, checking multiple files at once.  We'll need to think about making you print out two maps.
:robot: :+1:

Issue #126, opened 2020-09-02
gBOpen - MDV ADM 2


slfuhrig:
We may have a raw file of ADM 2 for the Maldives that needs to be processed. If not, we'll need to search for a new one.

slfuhrig:
Check the raw folder first to see if there is an ADM 2 shapefile with an open license there. If not, we'll need to find a new file.

slfuhrig:
It looks like we need to find a new file, as no suitable candidates are listed in our raw data.

leeberryman:
I have a candidate.  So ADM2 would be considered the islands and ADM1 is the atolls correct?

PR #713, opened 2020-11-15
MDV_ADM2 #126

neelsimpson:
Cleaned MDV_ADM2 on 11/14

## Checklist:


slfuhrig (changes requested):
Here's a few things to fix to see if that'll help you pass the bot.
1. for the license png, please screenshot the metadata tab instead of the data tab
2. please save meta as a .txt file instead of whatever format it's in
3. you have to keep the "Other Notes:" field in metadata, even if you don't fill it
4. make sure to export your data as a shapefile
5. in the attribute table, only include one "Name" field - I'm not sure which one is correct here, so you may need to do some research.

PR #719, opened 2020-11-16
NPL_ADM2

helenabuckman:
Metadata validation

## Checklist:

- [X] I have included a file named "meta.txt", with all required metadata (see an example in https://bit.ly/2GtQ01i).


slfuhrig (changes requested):
SO close!! Even though the file is technically authoritative (since it comes from a gov site), we can use it for our humanitarian release due to the license type. So, change the release type in the metadata to "gbHumanitarian," and you should pass the bot. No manual review will be required because I checked everything (and it's all good to go).

slfuhrig:
Will submit new PR

PR #750, opened 2020-12-18
STP_ADM2

dorianvmiller:
metadata validation for STP_ADM2

 #440 

## Checklist:


nmspage (changes requested):
Along with the "Name" field, it looks as though there's a visible discrepancy between ADM2 and ADM0, an island to the south is missing from this shapefile. Easy fix though! Could you take the ADM0 layer, use the "explode" and add the island to ADM2? Should be easy to look up the name of the island. Let me know if you have questions!

dorianvmiller:
Good catch! I added the island from the ADM0 layer - turns out it's a part of one of the districts already in the ADM2 layer so I merged them together (Caué in the attribute table). Let me know if anything else needs to be changed!

nmspage (approved):
Awesome thanks!!

nmspage (changes requested):
My mistake in forgetting this earlier - I realized that we need to change source 1 in the metadata to geoBoundaries because of the edits. Please also add that there were geometry changes in the notes. Source 2 becomes the old source 1. Thank you!

nmspage (approved):
Perfect, thank you!

DanRunfola:
Thank you!

Issue #56, opened 2020-07-15
gB OPEN - Kenya ADM3


DanRunfola:
**"Creative Commons Attribution - NonCommercial 4.0 International (CC BY-NC 4.0)"**
Kenya ADM3  This has been redefined as a "Humanitarian" license, and is no longer in the Open Release.  

Next step is to find a version we can keep in the open release.

PR #783, opened 2021-01-08
KEN_ADM3 gbOPEN #56

nmspage:
 #56 

## Checklist:


slfuhrig (changes requested):
So close! Just need some simple fixes. Please list "Humdata" as Source 1. Also, do we have "CC Attribution International" in the list of acceptable licenses (aka is the license here 4.0 or no?)?

slfuhrig:
Continues in #784

PR #807, opened 2021-01-14
MKD_ADM1 Metadata Validation

kpupkiewicz:
#414 

## Checklist:


slfuhrig (changes requested):
Please add "ISO_Code" field to attribute table.

slfuhrig (changes requested):
I assumed there would be ADM1-specific ISO_Codes but don't see any after looking for a bit, so apologies, but please remove the field again since we can only do the country-wide ISO or a more specific one.

slfuhrig (approved):

PR #839, opened 2021-02-08
ECU ADM2 gbOpen #491

carolinamonrivera:
https://github.com/wmgeolab/gbRelease/issues/491

## Checklist:


sylviashea1 (changes requested):
Hey Caro! There's a problem with your attribute table. It includes several unneeded columns, and the name field(s) are glitchy - they include numbers, strange spacing, and a bunch of blanks. Otherwise, everything else is good.
Screen Shot 2021-02-08 at 9 34 00 AM

slfuhrig:
Since source 1 for this file is ipums.org, we cannot use it (their license isn't compatible with our release), so we will need a new file from an open license source.

Issue #156, opened 2020-09-11
gBOPEN PHL ADM 3


slfuhrig:
Our existing ADM is authoritative and humanitarian.

DanRunfola:
To clarify - we need to find a new gbOpen for this.

lydiatroup:
Found this shapefile: https://web.archive.org/web/20190618021200/http://philgis.org/country-barangay/country-barangays-file

But the website no longer exists, and unknown usage information

PR #861, opened 2021-02-15
PHL_ADM3 gbOpen

nmspage:
 #156 

## Checklist:


slfuhrig (approved):

DanRunfola:
So, if I accept this, we would be deleting Poland and Ireland :))).

I need to add some bot checks for "deleted files".

DanRunfola (changes requested):
This is deleting Ireland and Poland :).

Might be easiest to just submit a new PR with only the one file.

Issue #824, opened 2021-01-26
gB OPEN -- NZL ADM4


slfuhrig:
Fire & Emergency New Zealand recently released a Localities dataset which may serve as ADM4 for NZL.
https://data.linz.govt.nz/layer/104830-fire-and-emergency-nz-localities/

We need to double check to make sure the data is compatible then upload it if so.

rcoup:
@palmerj this & #883 something your team could advise on / help with?

DanRunfola:
Closing stale issues for ADM3 and below.

DanRunfola:
Thank you for flagging - reopening and tagging!

PR #883, opened 2021-02-19
NZL_ADM4 gbOpen

nmspage:
 #824 

## Checklist:

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -28,6 +28,7 @@ sourceData/gbOpen/NZL_ADM0.zip filter=lfs diff=lfs merge=lfs -text
 sourceData/gbOpen/NZL_ADM1.zip filter=lfs diff=lfs merge=lfs -text
 sourceData/gbOpen/NZL_ADM2.zip filter=lfs diff=lfs merge=lfs -text
 sourceData/gbOpen/NZL_ADM3.zip filter=lfs diff=lfs merge=lfs -text
+sourceData/gbOpen/NZL_ADM4.zip filter=lfs diff=lfs merge=lfs -text
 sourceData/gbOpen/PAN_ADM3.zip filter=lfs diff=lfs merge=lfs -text
 sourceData/gbOpen/POL_ADM3.zip filter=lfs diff=lfs merge=lfs -text
 sourceData/gbOpen/PRT_ADM3.zip filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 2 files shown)


slfuhrig:
There is an issue with the topology of this boundary that needs to be addressed due to the presence of hundreds of small polygons for every lake in the country.

larogersWM:
This layer is filled with lots (900+) little lakes and Inland Bays as individual polygons that need to be fixed. There are tiny lakes completely surrounded by a suburb/localities, larger lakes that span multiple localities, and inland bays that are even larger, which makes removing the lakes even more difficult. I'm wondering if there is a way to dissolve polygons completely surrounded by one other polygon. I left the attribute table as given, just in case.

DanRunfola:
Thanks for posting this PR - this is a unique one, and really challenges our definition of "terrestrial".

A little background:
1) geoBoundaries focuses on land borders - i.e., we don't handle maritime boundaries.
2) However, we don't (as an example) remove rivers or streams from our data; more generally, landlocked lakes (well, or lakes that are fully surrounded by land within a single country) are retained, but we've never really made a firm decision on that.

In this case, we're faced with a few unique issues:
1) The lakes are individual polygons, so simply removing them would result in a huge number of holes in the data.
2) We don't know the appropriate way to dissolve lakes, so as to ascribe ownership to the appropriate polygon.
3) We can't leave the lakes, as our database only has administrative boundaries - i.e., each row is an adm boundary.

I'm leaning towards - for now - "remove all the lakes and leave the holes".  We then create an issue for ~gb 5.0 to go in and manually correct it by aggregating the lakes with the appropriate boundaries.  Recognizing it's a bit of a bummer to go that way, but I can't think of a better solution.

What do ya'll think?

slfuhrig:
If we cannot find a suitable shapefile without lakes from another source, I agree that it's best to just remove them from this file for the 4.0 release and work to find another for 5.0 My concern is that the problem exists in ADM0-3 as well.

larogersWM:
I just checked NZL_ADM 1, 2, and 3 and the lakes are not a problem there because the shapefile comes from a different source. I I have figured out how to manually select the lakes and fill them, which would help for the very tiny lakes that we know which suburb/locality they belong in (i.e. if the lake is completely surrounded by one suburb/locality). Then I can leave bigger holes for the large lakes and inland bays. This will be time consuming, so if it not the most important or will likely not result in the layer we want, I can simply remove the lakes and leave a lot of holes in the layer for 5.0 -- I am fine with either.

DanRunfola:
I'm goign to go ahead and accept this one - our auto-gap-filling algorithms will catch very small lakes.  We'll want to revisit in the future for a more permanent solution.

DanRunfola:
Oh, well, I'll accept after we fix: 
Detected Key / Value: Boundary Type / Suburbs and Localities
CRITICAL ERROR: The boundary type in the meta.txt file is invalid: Suburbs and Localities
We expect one of: ['ADM0', 'ADM1', 'ADM2', 'ADM3', 'ADM4', 'ADM5']

Issue #110, opened 2020-08-13
gbOPEN - NER - ADM2


nmspage:
NER ADM2 was moved to humanitarian release. Need to find an alternative source for gbOPEN product.

lydiatroup:
Found a map to digitize here: http://www.citypopulation.de/en/niger/admin/?mode=status&map=osm_dlr&opacity=0.8 
Is Creative Commons 3.0.

Has 63 departments (ADM2 name) as of 2011

DanRunfola:
Closed by https://github.com/wmgeolab/geoBoundaries/pull/888

PR #886, opened 2021-02-20
NER_ADM2

lydiatroup:
digitized file of NER adm2, addressing issue #110

https://github.com/wmgeolab/gbRelease/issues/110

## Checklist:


slfuhrig (changes requested):
We just need a few small changes:
1. the name of the license screenshot has ".png" in it twice, so that's failing the bot -- the name can either be "license" or "license.png"
2. for release type in meta, it needs to be "gbOpen" because the bot is super specific
3. change the license type to "Creative Commons Attribution 3.0 License" to match what the bot will recognize (list linked in drive directory for future reference)
4. in meta, you need to have all of the fields included even if they don't have any info in them (aka if optional and not in use leave blank) -- you need to add "source 2" back in 
5. since this is a digitized adm, our source 1 is technically "geoBoundaries" and source 2 is what you currently have listed as source 1
6. remove the ISO_Code field from the attribute table since this is ADM2 (we only do ISO for ADM0-1)

I think that's everything. I'll double check when you resubmit. Lemme know if you have any questions :)

PR #924, opened 2021-03-02
CAF ADM3

charliealtman:
CAF ADM3 are the city districts of the capital city, Bangui. In the shapefile, there are also polygons labeled "Bimbo." I explained what this means in the meta.txt file

 CAF_ADM3

 #533 (Also in response to discovering CAF ADM4 was really CAF ADM 5 and I could dissolve boundaries to get ADM3)

## Anything Unusual?
 see meta.txt file
 ^^^

## Checklist:


slfuhrig (changes requested):
Was this dissolved from another ADM level? If so, "geoBoundaries" needs to be listed as source 1.

PR #960, opened 2021-03-06
CHE_ADM3 gbOpen

rmbiddle:
 #935

## Anything Unusual?

 No complications but downloaded from OSM, looks like there are some districts missing (big gap in the middle of the country)

## Checklist:


slfuhrig (changes requested):
There's a large hole in the center of the shapefile. Moving forward we need to find out if there is actually a gap in ADM coverage (if so link sources), or if it's a problem with this specific shapefile.

Issue #87, opened 2020-08-13
gbOPEN - GMB - ADM0


nmspage:
GMB ADM0 was moved to humanitarian release. Need to find an alternative source for gbOPEN product.

hadleyday:
Notes on Gambia (before break starts) 


https://www.researchgate.net/figure/Map-of-The-Gambia-showing-administrative-boundaries-population-density-and-districts_fig9_283775741

https://sedac.ciesin.columbia.edu/data/set/gpw-v3-subnational-admin-boundaries/maps?facets=region:africa

https://wascal-dataportal.org/2.0/?q=dataset/administrative-regions-gambia

PR #977, opened 2021-03-10
issue #87 GMB ADM_0

hadleyday:
## Checklist:


slfuhrig (changes requested):
The bot is VERY picky -- I think that's why you're failing some of the checks. First, the name of the file you uploaded is "GMB_ADM0 (1)." Make sure the name is only "GMB_ADM0" for both the zipfile and shapefile inside. The license screenshot also has ".png" twice, so I'd remove that from the name so it's only listed once. For the metadata: boundary type should just be "ADM3", canonical name whatever the local name for GMB is, the license name needs to match the list of acceptable licenses exactly (it's easiest to just copy and paste from the list into metadata). Also, the fields in the attribute table need to be capitalized. If you need help with any of the changes, pop into office hours :) Once you make those changes, reupload and let's see if you pass.

sylviashea1:
Fixed with PR #990

PR #984, opened 2021-03-15
BGD_ADM2 humanitarian #941

slfuhrig:
 #941 

## Checklist:


sgmurphy00 (changes requested):
Metadata license type does not concur with license picture- should be Creative Commons Attribution for Intergovernmental Organisations.

sgmurphy216 (changes requested):
Looks like you submitted two copies! Please resubmit 'BGD_ADM2' .zip and we'll be good to go! Thanks Sydney!

PR #985, opened 2021-03-15
#308 syr_adm1 humanitarian

slfuhrig:
 #308 

## Checklist:


sgmurphy216 (changes requested):
Hey Sydney! Looks like this would fit into gB Open via CC3.0 IGO; this may be why the meta check failed. Also, boundary representative of year appears to be '21 in the license png, so please correct that. Thanks!

Issue #106, opened 2020-08-13
gbOPEN - MOZ - ADM2


nmspage:
MOZ ADM2 was moved to humanitarian release. Need to find an alternative source for gbOPEN product.

sylviashea1:
Assigned to Hadley Day (not in gB Org yet)

hadleyday:
Mozambique notes (before break)
Mozambique 2020 Updated Districts List 
Country Breakdown
Mozambique (ADM 0)
10 Provinces and 1 capital city (ADM 1)
128 Districts (ADM 2) 
405 Postos (ADM 3) 

Possible lead:
https://catalog.loc.gov/vwebv/staffView?searchId=1421&recPointer=0&recCount=25&searchType=1&bibId=16968032
Humanitarian Use Links:
https://data.humdata.org/dataset/mozambique-administrative-levels-0-3

PR #989, opened 2021-03-16
Add files via upload

taraamcl:
MOZ ADM2 self-assigned #106
one note is that wikipedia says there are 129 districts but the HumData file had around 150 districts


slfuhrig (approved):

slfuhrig:
It's fine that there is a difference in numbers reported by Wikipedia and Humdata. In this case, we defer to Humdata since it has UN info and is a more reputable source.

DanRunfola:
Thank you!

Given the uncertainty here, I did want to add a wiki page, so I did!:
https://github.com/wmgeolab/geoBoundaries/wiki/MOZ

PR #1006, opened 2021-03-21
Update metaSTPLNEXFUN.txt

owachob:
Unsure of if that would be the correct unit, old source link was dead so not sure if this is it's replacement or not.


coyotefarrell (changes requested):
Typo of license

Issue #88, opened 2020-08-13
gbOPEN - GMB - ADM1


nmspage:
GMB ADM1 was moved to humanitarian release. Need to find an alternative source for gbOPEN product.

dfornatora:
Trouble finding open license source. Have found that ADM1 is split into 5 Regions and 1 City.

lydiatroup:
see here for a map to digitize: https://en.wikipedia.org/wiki/Subdivisions_of_the_Gambia

PR #1017, opened 2021-03-24
GMB_ADM1 gbOpen

rmbiddle:
Digitization of GMB_ADM1

 #88


slfuhrig (changes requested):
change boundary type in meta to "ADM1"

slfuhrig (changes requested):
So close! If there are specific ADM1 ISO codes, please add a field for them to be included.

slfuhrig:
please submit a new PR for this issue - there are duplicate files here, so that'll be easiest.

Issue #89, opened 2020-08-13
gbOPEN - GMB - ADM2


nmspage:
GMB ADM2 was moved to humanitarian release. Need to find an alternative source for gbOPEN product.

lydiatroup:
see here for a map to digitize https://en.wikipedia.org/wiki/Districts_of_the_Gambia 
could be helpful to do GMB1 first, then use those files & the maps on this page to digitize & split up a adm1 file into adm2s

PR #1018, opened 2021-03-24
GMB_ADM2 gbOpen

rmbiddle:
GMB_ADM2 digitization

 #89


slfuhrig (changes requested):
same comment as for ADM 1 - change boundary type to "ADM2"

slfuhrig (changes requested):
So close! Source 2 in meta is Wikipedia since that's where the jpeg came from.

slfuhrig:
You accidentally put one file in the root instead of in source data > gB open, which means the checks won't work. I'm going to close this PR. Please create a new PR with the edited file in the correct location :)

PR #1020, opened 2021-03-25
VEN_ADM1 gbOpen

rmbiddle:
 #424


slfuhrig (changes requested):
Everything is perfect!! We just need ISO_Code field in the attribute table since this is ADM1. 

https://www.iso.org/obp/ui/#iso:code:3166:VE

PR #1040, opened 2021-03-28
Update metaTWOMINREAD.txt

owachob:
No working source link, was unsure of original data source and terms of use for new link used.


coyotefarrell (changes requested):
Typo of "liscence"

coyotefarrell (approved):

PR #1041, opened 2021-03-28
Update metaRORGPREEDU.txt

owachob:
Was unsure of if this link was correct for source link as the old one did not work.


coyotefarrell (changes requested):
Typo of "liscence"

Issue #64, opened 2020-07-15
gbOPEN - LBR ADM 0, 1, 2


DanRunfola:
LBR ADM 0, 1 and 2 need open licenses for gbOPEN ; current boundaries were moved to gb Humanitarian.

PR #1053, opened 2021-03-29
LBR_ADM1 gbOpen

rmbiddle:
 #64


slfuhrig (changes requested):
So close! We need unique ISO codes for each county. check iso.org

Issue #1062, opened 2021-03-31
gb open kenya adm 3


slfuhrig:
our current file is incorrect, so need a new one

PR #1063, opened 2021-03-31
KEN_ADM3

lydiatroup:
addressing issue #1062

 https://github.com/wmgeolab/geoBoundaries/issues/1062


slfuhrig (changes requested):
So close!! we need a "Name" field instead of "ward" in the attribute table, so please change :)

PR #1064, opened 2021-03-31
KEN_ADM3

lydiatroup:
addressing issue #307

 #307


slfuhrig (changes requested):
Again, we need a "Name" field in the attribute table :) otherwise perfect

PR #1070, opened 2021-04-01
MEX_ADM1 gbOpen

rmbiddle:
 #466


slfuhrig (changes requested):
So close! We need specific ISO codes for each state - check out iso.org

slfuhrig (approved):

DanRunfola:
Thank you!

Issue #880, opened 2021-02-18
IRN 2-4 holes - gb open


slfuhrig:
Iran's administrative hierarchy is divided into a system of provinces, counties, and further subdivisions. In ADM 0-3 layers there is a hole in the province of Tehran where the capital city of Tehran is located. The ADM 3 layer has other holes as well: these appear to be centered on the cities of Isfahan, Dehdasht, Mashhad. The ADM 4 layer maintains some of these holes present in the ADM 3 layer while containing other holes, such as those around Malayer.

These discrepancies present an interesting dilemma moving forward. There are two likely possibilities:

These boundaries are accurate and represent Iran's true administrative divisions.
These boundaries contain inaccuracies at some level.
Moving forward, more research will need to be conducted to discern the true nature of these holes. If these boundaries are indeed inaccurate, they will be replaced.

larogersWM:
Running into issues with IRN_ADM4. IRN_ADM4 has holes in the layer that must be filled. The problem is these holes are holes of ADM3 polygons and are not further divided into the proper ADM4 polygons. (screenshots below). There are no maps of ADM4 to digitize, so I cannot divide the polygons myself. Any suggestions?

ADM3:
![image](https://user-images.githubusercontent.com/70722024/111820277-ae530080-88b7-11eb-81c5-553446796047.png)

ADM4:
![image](https://user-images.githubusercontent.com/70722024/111820075-71870980-88b7-11eb-8513-71b6c2b9e9df.png)

The red outline represents the hole in the layer, which is the same between the two images, but there more lines (boundaries) in the ADM4 layer. The ADM3 polygon should be further divided but is not.

PR #1073, opened 2021-04-02
IRN_ADM2 and IRN_ADM3 #880

larogersWM:
Fixed the holes in IRN_ADM2 and IRN_ADM3. I do not have enough information to fix the holes for IRN_ADM4


slfuhrig (changes requested):
so close! you forgot canonical boundary name for both.

slfuhrig (approved):

PR #1086, opened 2021-04-04
Update metaNOPARISPRI.txt

owachob:
Changed source link, went with the units as listed in the data set in drive, could not find license(consistent with magic tracker


coyotefarrell (commented):
Is it presented as a number or 0 or 1?

jcsiwek (changes requested):
rc for source

jcsiwek (approved):

PR #1091, opened 2021-04-05
Update metaLRPLTRECON.txt

nirajpatel10122:
Changed the data definition, units, data collection methodology, license, license source, link to source data.


coyotefarrell (changes requested):
Can you put most of the information in Unit under Data Definition and keep unit to something very simple like number or percent?

jcsiwek (approved):

Issue #73, opened 2020-08-13
gbOPEN - BGD - ADM3


nmspage:
BGD ADM3 was moved to the humanitarian release. Need to find an alternative for the gbOPEN product.

DanRunfola:
Was duplicated in #933 .  Closing that one and leaving this one open. Bot may get aggresive and try to recreate, copying this here to try and preclude:

====robotid-d7329e7104s40t927830R028o9327y372h87u910m197a9472n2837s649====

taraamcl:
these are "upzilas" but it has been a difficult boundary to find. There are around ~500 of these depending on what source you look at.

PR #1114, opened 2021-04-09
BGD_ADM3 gbOpen

nmspage:
This is @taraamcl 's PR

  

 #73


slfuhrig (changes requested):
It's pretty close! The only problem is that the meta appears to be for LSO. So, please double check all the fields and correct where necessary before submitting again.

slfuhrig (approved):

DanRunfola:
Thank you!!

Issue #1131, opened 2021-04-12
IRL_ADM2 gbOpen


nmspage:
The current IRL_ADM2 that we had was actually ADM3, need to find a new ADM2 for gbOpen.

nmspage:
Upon further research (see https://github.com/wmgeolab/geoBoundaries/wiki/IRL) we have decided to include counties as IRL ADM1 and townlands as IRL ADM2.

PR #1143, opened 2021-04-16
IRL_ADM2 gbOpen

rmbiddle:
 #1131


sgmurphy216 (changes requested):
Hey Russ! Geometry looks fine, but from my research, ADM2 should be a 'county' level boundary, consisting of 26 counties in IRL, followed by municipals ADM3. It may be that our current ADM3 actually corresponds to 'townships/parishes' @ ADM4; more research required there.

Issue #1061, opened 2021-03-31
gb open kenya adm 2


slfuhrig:
needs to be updated, what we have is incorrect

PR #1148, opened 2021-04-16
KEN_ADM2

lydiatroup:
KEN_ADM2 open to address issue #1061

https://github.com/wmgeolab/geoBoundaries/issues/1061

 https://github.com/wmgeolab/geoBoundaries/issues/1061


sgmurphy216 (approved):
Hey Lydia! Looks like a pretty good source, but there is one geometric error in Kanduyi (ring self-intersect). Gonna go ahead and approve this and see what the build script thinks, but this may require another look.

[KEN_ADM2_GeometryError.pdf](https://github.com/wmgeolab/geoBoundaries/files/6345233/KEN_ADM2_GeometryError.pdf)

DanRunfola:
We should be able to catch some of the geom errors on the back-end, but definitely good to start flagging them.

Issue #777, opened 2021-01-07
gB OPEN -- CHL ADM3 


slfuhrig:
We need CHL ADM 3 (communes).

taraamcl:
uploaded to drive for natalie to PR

PR #1154, opened 2021-04-17
CHL_ADM0

taraamcl:
CHL ADM0 #777


lydiatroup (changes requested):
metadata and license look good, but it looks like the shapefile doesn't have any features in it (nothing comes up in arc/q when you add it). 
Try exporting it again and make sure nothing is selected! To check if it works try adding the new file into your map and see if the shapefile shows up

lydiatroup:
was uploaded with lfs - closing this issue

Issue #50, opened 2020-06-30
geoBoundaries OPEN - SALB Licensure Issue - Gabon


DanRunfola:
Gabon (from SALB) has a license issue we will need to either correct or fall back to an alternative source:

- SALB released a validated version of the 2nd level admin level for the January 2000-January 2006 period during its first phase of activity (2001-2011). This is the layer is the one accessible from the HDX
- The layer integrated into GeoBoundaries has itself been adjusted to correspond to the situation observed in the country since February 2013 

Because the original dataset was modified, and we release standardized international boundaries, we will likely need to fallback to an alternative use.  In this case, we also may consider an upstream contribution to SALB if we can get permission.

lydiatroup:
https://data.humdata.org/dataset/gabon-administrative-boundaries 

was updated 2 months ago? may be fixed? look into

if not working: this map can be digitized: https://en.wikipedia.org/wiki/Departments_of_Gabon#/media/File:GabonDepartments.svg

PR #1155, opened 2021-04-17
GAB_ADM0,1,2 gbOpen

rmbiddle:
Digitization of GAB_ADM0-2

 #50


lydiatroup (changes requested):
thank you for doing these so quickly! a few things need to be fixed and then they'll be good to go. 
1) one of your ADM2 features has "NULL" in the attribute table where it should say adm2, you need to edit that file so that one says ADM2 too
2) for ADM1 and 0, because you used a different file (from 2005) you need to update the year on your metadata to 2005 instead of 2017
3)  i'm pretty sure you don't need to include the image you digitized in your zipfile, but am checking with sydney and silvia so hold off on resubmitting these files until i confirm with them and let you know

lydiatroup (changes requested):
just a couple things left to fix!
1) projection: when you digitize a boundary it defaults to use whatever projection your map was in, which is generally not the one we want. To fix this, you'll need to project your digitized shapefiles into the correct coordinate system ( EPSG 4326 ) 
2) your adm1 zipfile has another zipfile inside of it that you'll need to remove
3) there's still that null value for one of your ADM2's level field. To fix this you can either edit that specific field for that specific feature or you could make a new field for level and set them all to = adm2.  
let me know if you have any questions!

sgmurphy216 (approved):
Great work Russ- super granular digitization! Closes #50

lydiatroup (approved):

DanRunfola:
Wow, wonderful detail on this.  Nice work.

PR #1177, opened 2021-04-19
TWN ADM0#1170

slfuhrig:
 #1170


sgmurphy216 (changes requested):
Hey Sydney! Looks like you accidentally zipped a folder in a folder; please re-zip meta, license, and .shp data and resubmit, we should be all good!

DanRunfola:
Triggering a re-run, just added TWN to the bots list of acceptable ISO codes.

DanRunfola:
This is still failing because of the zip folder structure, FYI.  Need to fix that for the zipfile check and meta checks to pass.

PR #1179, opened 2021-04-20
GIN_ADM3

charliealtman:
You're welcome!
 GIN_ADM3

 bug resubmission


sgmurphy216 (changes requested):
Hey Charlie! Shapefile looks great, please just change Source 1 in meta.txt to correspond with the 'source' field in the license png. Thank you!

PR #1226, opened 2021-04-22
Issue #1171 TWN_ADM1

hadleyday:


lydiatroup (changes requested):
A couple changes are needed and then it'll be good!
1) your attribute table has an empty "Level" field, and a "Level_1" field with the level in it. (only need 1 "Level" with the level)
2) the meta issue with dan lol

lydiatroup (changes requested):
your zip file has the folder with all the stuff titled "TWN_ADM1" and a "mac os" file so it's failing the meta checks since the bot can't access the metadata (bc it's in a subfolder). 
also your shapefile has 2 fields for level, 'level' which is empty and 'level_1' which has the adm # please fix these issues and resubmit!

lydiatroup (changes requested):
still a couple issues left:
1) there's an empty folder called 'new folder' in your zipfile, please remove this
2) there's still two columns for level (level which is null and level_1 which has the correct values)
please fix these and resubmit!

lydiatroup (approved):
looks great! thanks hadley! Closes issue #1171

DanRunfola:
Glad to see this one made it through :)  Thank you!

Issue #59, opened 2020-07-15
gb OPEN - URY ADM 2


DanRunfola:
URY ADM 2 is currently non-commercial. As a part of our move towards a more unified product, we need to find an alternative.

The existing URY ADM2 has been moved into our forthcoming Humanitarian release, which allows for more restrictive licensing.

SidonieH:
It looks like we already have this as open? From OSM?

dorianvmiller:
Resolved by #1583

PR #1249, opened 2021-04-23
#59 URY_ADM2

SidonieH:
Needed to find new open data. From OSM

  #59


sgmurphy216 (changes requested):
Hey Sidonie! This source may not be great for us; geometry looks to be off, and the fields aren't consistent with our conventions. Probably worth finding a new source for URY2

Issue #1213, opened 2021-04-20
gB OPEN SLV_ADM1 


sgmurphy216:
Failed gB OPEN build script; appears to be missing metadata. Please located metadata and resubmit.

SidonieH:
The file uploaded to sourceData gbOpen appears to have correct metadata in the zip. Not sure what to fix

nmspage:
This issue was closed with PR #1027 and approved by Sydney.

PR #1252, opened 2021-04-24
#1213 SLV_ADM1

SidonieH:
 #1213


lydiatroup (changes requested):
It looks like you forgot to add the file to your pull request! Please edit this request and add the file!

lydiatroup (changes requested):
it still looks like there's no files uploaded. if you're doing the normal upload methodology and it's still not working, i'd say try an LFS submission and hopefully that will work for us

Issue #696, opened 2020-11-13
gB Authoritative CHL_ADM0


slfuhrig:
Our file in gB Open is a candidate, just need to update meta and re-upload here.

taraamcl:
I put this on drive for natalie to upload

PR #1255, opened 2021-04-24
CHL_ADM0 gbAuthoritative

nmspage:
 #696


nmspage:
This is @taraamcl 's PR

sgmurphy216 (changes requested):
Hey Natalie/Tara! Few changes to be made here: please resubmit a .zip with all supporting files (missing .prj; just re-export the data from arc or q and re-zip, should be fine). Also, in meta.txt, you put 'gb open' as release type, this should be authoritative. Finally, just check your projection and we should be good to go! Thanks!

sgmurphy216 (approved):
Looks great! Thanks! Closes #696

DanRunfola:
Thank you!

Issue #1169, opened 2021-04-19
[BOUNDARY ERRATA] AUS ADM2


DanRunfola:
**Describe the Error**
AUS ADM2 has a topological error we can't algorithmically fix

**Screenshots**
![image](https://user-images.githubusercontent.com/7882645/115287201-fadf6500-a11d-11eb-85cd-71b7b2172127.png)

There is a small gap between boundaries located roughly at where the circle exists (thanks to Max at the WFP for putting together this image).

hadleyday:
https://github.com/wmgeolab/geoBoundaries/pull/1581 - did not link the issue to the PR

PR #1289, opened 2021-04-29
gB Open Aus ADM2

SidonieH:
 #1169


sylviashea1:
Error with layer, do not approve. Will be re-uploaded

sylviashea1 (changes requested):
For reference - subpoly of AUS uploaded, not whole shapefile.

Issue #1214, opened 2021-04-20
gB OPEN SLV_ADM0


sgmurphy216:
Failed gB OPEN build script; appears to be missing metadata. Please located metadata and resubmit.

PR #1303, opened 2021-05-02
gbOpen SLV_ADM0

carolinamonrivera:
Issue #1214

https://github.com/wmgeolab/geoBoundaries/issues/1214

## Anything Unusual?

Need to add this license to list "Attribution 2.5 Denmark (CC BY 2.5 DK)"


slfuhrig (changes requested):
Please address failing checks -- also with the shapefile itself, it looks like a bunch of different pixels instead of one large shapefile. Ideally, please find a file that isn't comprised of pixels (to get more accurate borders), even if that means dissolving a higher level ADM.

lydiatroup:
closing this bc we just approved another pull request for the same issue!

Issue #1145, opened 2021-04-16
gb OPEN CHE_ADM3


slfuhrig:
We need a new shapefile

One lead: edit WMS link: http://wms.geo.admin.ch/?SERVICE=WMS&VERSION=1.3.0&REQUEST=GetCapabilities&lang=en

PR #1320, opened 2021-05-06
CHE ADM3

memulder:
issue #1145. Multipart polygons for municipalities were dissolved based on the name field.

 #1145


slfuhrig (approved):

DanRunfola:
Thank you maddy!!!

PR #1321, opened 2021-05-06
gbOpen BGD_ADM5

leeberryman:


sgmurphy216 (commented):
Thank you for the contribution! On visual inspection, there may be errors with attributes, and it looks like there are a number of geometric errors primarily in the southeastern region of BGD; many boundaries extend over water, and in areas, borders appear to be inconsistent with current gB files/OSM basemap. Unfortunately, we will not be able to accept this submission.

PR #1331, opened 2021-05-08
SAU_ADM2 #510

larogersWM:


lydiatroup (changes requested):
both the Name and Level fields have ADM2 as the values. other than that, looks good!

PR #1343, opened 2021-05-24
XKX_ADM2 gbOpen

carolinamonrivera:
issue #1175

seems like original file in gbOpen was missing a license.png, did not notice anything wrong with field names

XKX_ADM2

https://github.com/wmgeolab/geoBoundaries/issues/1175

## Anything Unusual?
noticed that overall shp contained 8 files instead of 6 -- maybe could be signal for failed script?


sgmurphy216 (approved):
Looking great! Thanks for the fix Caro! Closes #1175

Issue #1190, opened 2021-04-20
gB OPEN UKR_ADM2


sgmurphy216:
gB OPEN build script failed; appears meta.txt not included in submission. Please include and resubmit.

charliealtman:
Update 1: Ukraine drastically changed their ADM2 last July (2020) most files I've found on the internet so far do not reflect those changes.

PR #1345, opened 2021-05-24
UKR ADM2 Issue #1190

dfornatora:
 https://github.com/wmgeolab/geoBoundaries/issues/1190


sgmurphy216 (changes requested):
One small change- please replace 'UKR' in Level field with 'ADM2' and we should be good to go!

sgmurphy216 (changes requested):
Looks like this source may be missing the region containing Kiev as well as a central northern ADM2 comprising part of the exterior border. Any other sources we can look at? 

Edit: My mistake! This is canonically correct under UKR_ADM2's organization, these areas are not represented intentionally. Thanks Dominic, this looks good.

sgmurphy216 (approved):

DanRunfola:
@dfornatora @sgmurphy216  Could ya'll add a note to the wiki on the northern ADMs?

Issue #69, opened 2020-07-15
gbOPEN - AGO - ADM 3


DanRunfola:
AGO ADM 3 has been moved into our Humanitarian release; we will need to find an alternative for the gbOPEN release.

PR #1348, opened 2021-05-24
AGO_ADM3 gbOpen

nmspage:
 #69 

## Anything Unusual?

 This file was taken from HDX, and the file contributors are GADM and Instituto Nacional de Estatistica (INE). The file was edited by OCHA based on census data from INE. Physical edits were made to the  boundaries to ensure topology accuracy across ADM levels and to the Catumbela and Cacula municipalities. Considering that these physical edits, along with other boundary changes by digitizing, were made, is it an acceptable submission to the database?


DanRunfola:
Because this dataset includes GADM as one of it's components, we should probably avoid inclusion at this time in gbOpen.  Definitely works with gbHumanitarian, though!

Issue #182, opened 2020-09-25
gbOPEN NRU ADM 1


sylviashea1:
NRU_ADM1 file in gbOpen/master is incomplete. Must be replaced with accurate file.
Action: either revise existing shapefile or find a new one

PR #1350, opened 2021-05-25
Issue #182 NRU_ADM1

hadleyday:
NRU ADM1


sgmurphy216:
Per meeting- ADM1 constituencies, 8 features, is in digi process, followed by districts, 14 features.

PR #1351, opened 2021-05-25
Issue #629 NER_ADM1

hadleyday:
NER ADM 1

 https://github.com/wmgeolab/geoBoundaries/issues/629


sgmurphy216 (approved):
Looks great!! Thank you! Closes #629

Issue #1238, opened 2021-04-23
gB Open ZWE_ADM3


lydiatroup:
Current file for ZWE ADM3 has numbers listed in the name column for features, we need to find the names / a file with names and upload

carolinamonrivera:
Unsure if under the ward/name column it should be numbers. There area over 1200 wards and I think they are mainly recognized as numbers, but wikiped. says there are names, but every single shpfile online has them listed as numbers...

PR #1382, opened 2021-05-28
Zimbabwe ADM3 Issue #1238

MichaelR-geoLab:
ZWE shapefile with metadata and license


sgmurphy216 (approved):
Great fix Michael! Thanks! Closes #1238

Issue #1365, opened 2021-05-27
gB OPEN SLE_ADM1


sgmurphy216:
SLE_ADM1 failed build script; appears meta.txt was not included in submission. Please resubmit with meta.txt.

* Issue may have arisen during 3.0 transition from google drive hosting to git. If unable to find source files on github, check the drive tracker / www.geoboundaries.org for metadata.

DanRunfola:
Closed by https://github.com/wmgeolab/geoBoundaries/pull/1424

PR #1387, opened 2021-05-28
SLE_ADM1 issue #1365

rsritchey:


sgmurphy216 (changes requested):
Hey! Little typo in Source 1 field in meta.txt- correct this and we should be good to go!

Issue #1390, opened 2021-05-28
gB OPEN PSE_ADM2


sgmurphy216:
Corrected ADM2 Boundary; @hadleyday fixed.

PR #1391, opened 2021-05-28
Issue #1390 PSE_ADM2

hadleyday:
PSE_ADM2

 https://github.com/wmgeolab/geoBoundaries/issues/1390


sgmurphy216 (approved):
Thanks Hadley! Closes #1390

Issue #1361, opened 2021-05-27
gB OPEN SEN_ADM1


sgmurphy216:
SEN_ADM1 failed build script; appears submission did not include meta.txt. Please resubmit with meta.txt.

sgmurphy216:
* Issue may have arisen during 3.0 transition from google drive hosting to git. If unable to find source files on github, check the drive tracker / www.geoboundaries.org for metadata.

PR #1392, opened 2021-05-28
SEN_ADM1 Issue #1361

dfornatora:
 https://github.com/wmgeolab/geoBoundaries/issues/1361


sgmurphy216 (changes requested):
One small change- please replace the '_' in SN_KE (Kedougou) with a hyphen for consistency, then we should be good to go! Thank you!

sgmurphy216 (approved):

sgmurphy216:
Thanks!

PR #1394, opened 2021-05-28
SEN_ADM3 Issue #1363

dfornatora:
 https://github.com/wmgeolab/geoBoundaries/issues/1363


sgmurphy216 (changes requested):
Hey- quick question on this. Light research says SEN_ADM3 should have 133 features, some sources say 123, this file contains 121. Any chance HDX included any kind of 'readme' on why they include this 121? Would like to put SEN_ADM3 into the wiki with info as to why we represent 121- or, we find another source that we can back up with a source containing that info.

sgmurphy216 (approved):

Issue #1383, opened 2021-05-28
gB OPEN BGD_ADM2


sgmurphy216:
BGD_ADM2 failed build script; appears submission was placed in incorrect subdirectory. Please resubmit to 'gBOPEN' as opposed to humanitarian and we should be good to go!

PR #1395, opened 2021-05-28
BGD_ADM2 Issue #1383

dfornatora:
 https://github.com/wmgeolab/geoBoundaries/issues/1383


leeberryman:
@dfornatora This appears to be missing shapes for Kulna.  Is this a new source?

dfornatora:
> @dfornatora This appears to be missing shapes for Kulna. Is this a new source?

Thank you for pointing this out. This was not a new source, this new commit was placing BGD_ADM2 into the correct subdirectory. I will look into this for the missing shapes.

leeberryman:
@dfornatora 👍 I could help with it but I'm out for the next week.  If it's still pending I'll look.

dfornatora:
> @dfornatora 👍 I could help with it but I'm out for the next week. If it's still pending I'll look.

Thank you! New source was found that includes all 64 districts

sgmurphy216 (approved):
Looks good @dfornatora - thanks for the correction, and thanks @leeberryman for pointing that out! Closes #1383

PR #1401, opened 2021-05-31
gbOpen SLE_ADM3

carolinamonrivera:
Issue #1367 
so proud just did my first merge loll

https://github.com/wmgeolab/geoBoundaries/issues/1367


sgmurphy216 (changes requested):
Hey Caro! Great job on the merge- the file looks great! Through a bit of research, I found there should be [190 chiefdoms](https://web.archive.org/web/20180221142807/http://www.africareview.com/news/Sierra-Leone-unveils-new-geographical-map/979180-4041036-4x8dfaz/index.html), reflecting a change in division from 149 to 190 in 2017. Here, only 165 are represented. We may need to go through and re-merge based on an updated list of adm3s from 2017. I haven't found that list yet, but I'd be happy to look at this with you later (still a great job on the merge!)

sgmurphy216 (approved):
Thanks for working on this Caro! I just added the wiki page on this. Closes #1367

Issue #1284, opened 2021-04-28
Azerbaijan ADM1 vs ADM2


karimbahgat:
**Describe the Error**
The boundary issue is that Azerbaijan's ADM1 and ADM2 appear to refer to the same level. Some boundaries are different here or there, but I assume that's mostly due them being from different sources, and that the OSM-based one extends to the coastal boundary zone. Not sure if it's limited to the simplified version or applies to all versions. Is this intentional or just an oversight? 

**Screenshots**
Screenshots from the geoBoundaries website. ADM1:
![image](https://user-images.githubusercontent.com/6413369/116397402-d888ce00-a826-11eb-99cd-5a9efd66ff83.png)
And ADM2:
![image](https://user-images.githubusercontent.com/6413369/116397426-de7eaf00-a826-11eb-95db-75ee7766d134.png)

**Additional context**
Thought I would report this boundary issue/question through the official channel, so it's probably not going to be formatted correctly.

DanRunfola:
Let's try to take a look at this before 4.0.  Looks like an error on our part.

sgmurphy216:
Karim-

Great catch. Both files here represent the district level of Azerbaijan. Preliminary research has this level either at 1 or 2; depending on if the distinction between Contiguous Azerbaijan and the Nakhchivan Autonomous Republic constitute ADM1. This is an error on our part, and some more work needs to be done researching before we fix, so I'm going to have us take a look at this in the coming weeks as we shift into summer work. Thanks for pointing this out!

PR #1402, opened 2021-06-01
Issue #1284 AZE_ADM1

hadleyday:
AZE_ADM1

 https://github.com/wmgeolab/geoBoundaries/issues/1284

## Anything Unusual?

 Not officially a division, but the differentiation makes geographical sense and ensures an accurate ADM2 layer.


sgmurphy216 (changes requested):
One small typo in 'other notes' meta.txt, then we're golden. Thanks for the work!

PR #1403, opened 2021-06-01
ESP_ADM4 gbOpen

nmspage:
## Anything Unusual?
 This is @MichaelR-geoLab 's work, issue number #137


sgmurphy216 (approved):
Manual inspection looks good! Thanks @MichaelR-geoLab and @nmspage ! Closes #137

DanRunfola:
mmm, there are so many ADM units that the bot seems to be crashing.  Going to work on this one before I accept it.

DanRunfola:
OK! 

image

Looks like there is at least one feature that is missing a geometry - i.e., it's a record only in the attribute table.  So, we need to find that before we can accept this boundary.

@nmspage @sgmurphy216

sgmurphy216:
Err- does not exist, municipal level = adm3

PR #1405, opened 2021-06-01
CHE_ADM2

dorianvmiller:
Fixed metadata issue - had to find new source for data. The "holes" in the country are there because not every ADM1 in Switzerland (Catons) is divided into ADM2 (Districts).

 #1381


sgmurphy216 (approved):
Looks good Dorian- ADM2 is covered in the wiki already, so we're golden. Thanks! Closes #1381

PR #1406, opened 2021-06-02
SLE_ADM1 issue #1365

rsritchey:


sgmurphy216 (changes requested):
Hey @rsritchey ! One quick correction; please change 'canonical boundary name' field to the name of the organizational level, in this case 'Province'. Other than that, great work!

Issue #1380, opened 2021-05-28
gB OPEN GIN_ADM3


sgmurphy216:
Pathing error in submission; gBOPEN/GIN/ADM3 as opposed to gBOPEN/GIN_ADM3. Please resubmit with correct pathing and we should be golden!

PR #1419, opened 2021-06-02
GIN_ADM3 Issue #1380

dfornatora:
 https://github.com/wmgeolab/geoBoundaries/issues/1380


sgmurphy216 (approved):
Looks great! Thank you! Closes #1380

PR #1420, opened 2021-06-02
Issue #1415 NOR_ADM0

hadleyday:
NOR_ADM0

 https://github.com/wmgeolab/geoBoundaries/issues/1415


sgmurphy216 (approved):
Thanks for the fix @hadleyday - hopefully this solves the issue. Closes #1415

Issue #1215, opened 2021-04-20
RUS-ADM0-3_0_0-B1


gabriellaIO:
### What is the expected behavior?
The extent of this shape in EPSG:4326 should not go beyond 180.0 degrees longitude. 

### What is the actual behavior?  The more detail the better.
The extent of the shape in QGIS is 

Extent | -180.0000000000000000, 41.1874296558497832 : 180.0000000000001421, 81.8612189000637045
-- | --

(Note that it is over 180.0 degrees) 

### Other notes on how to reproduce the issue?
Download the geoBoundariesCGAZ_ADM0.geojson, open the file in QGIS, select the RUS-ADM0-3_0_0-B1 polygon, export, "save selected feature as", check the extent of the new shape for Russian Federation. It will be the same as shown above. 

### Any possible solutions?
Use the Extract/clip by extent tool in QGIS to edit the shape to clip the longitude to 180.0 degrees.

PR #1425, opened 2021-06-02
RUS_ADM0

dorianvmiller:
Fixed extent to be within 180.00 degrees.

 #1215


sgmurphy216 (approved):
Great fix Dorian! Thanks! Closes #1215

DanRunfola:
Russia, always breaking my auto-map-makers.  *Shakes digital fist*

PR #1426, opened 2021-06-02
TUR_ADM2

dorianvmiller:
Old source not valid for gB Open, found new source and resubmitted with meta.txt

 #1371


sgmurphy216 (approved):
Looks solid- thanks! Closes #1371

PR #1423, opened 2021-06-02
Issue #1359 SDN_ADM2

MichaelR-geoLab:
uploading updated Sudan file with license and metadata

PR #1427, opened 2021-06-02
SDN_ADM2

dorianvmiller:
Old source not valid for gB Open, found new source and resubmitted with meta.txt

 #1359


sgmurphy216 (commented):
Issue completed by PR #1423

PR #1428, opened 2021-06-03
Issue #1284 AZE_ADM1

hadleyday:
AZE_ADM1

 https://github.com/wmgeolab/geoBoundaries/issues/1284


sgmurphy216 (approved):
Thanks Hadley! Closes #1284

Issue #1376, opened 2021-05-27
[BOUNDARY ERRATA] ADM0 for MMR includes water territory


leeberryman:
**Describe the Error**
Making issues for some data, not sure if this is already addressed for v4

PR #1431, opened 2021-06-03
Issue #1376 MMR_ADM0

hadleyday:
MMR_ADM0

 https://github.com/wmgeolab/geoBoundaries/issues/1376


sgmurphy216 (approved):
Thanks for the fix! Closes #1376

Issue #1413, opened 2021-06-02
gB OPEN NZL_ADM0


sgmurphy216:
NZL_ADM0 fails to load into geoPandas; please download file from www.geoboundaries.org , upload to www.mapshaper.org and simplify by 0.1% (99.9% on slider), download new file and reupload to Git. *Please make sure this preserves geometries, we aren't really trying to simplify here*

PR #1432, opened 2021-06-03
rmbiddle patch 47

rmbiddle:
 #1413


sgmurphy216 (changes requested):
Thanks Russ, NZL looks good! Looks like you also submitted LKA_ADM4 in this patch, which was submitted and accepted a while ago. Please resubmit with only NZL_0 and we're golden!

Issue #115, opened 2020-08-13
gbOPEN - LKA - ADM4


nmspage:
LKA ADM4 was moved to authoritative and humanitarian releases. Need to find an alternative for the gbOPEN product.

leeberryman:
I can help with LKA data.

PR #1433, opened 2021-06-03
LKA_ADM4 gbOpen

nmspage:
 #115


nmspage:
There were some issues with LKA being merged into the wrong branch and a personal repository in the past. This is a new file to resolve the issue.

sgmurphy216 (changes requested):
As discussed- please edit [unknown] name fields to 'Unknown 1' through 25 and create a wiki reflecting this change! Thank you!

sgmurphy216 (approved):
Thanks for the fix! Closes #115

Issue #1407, opened 2021-06-02
gB OPEN SDN_ADM0


sgmurphy216:
SDN_ADM0 is failing build script (geometric errors). Please find a new source and upload.

PR #1437, opened 2021-06-03
SDN_ADM0

dorianvmiller:
Found new source for SDN_ADM0.

 #1407


dorianvmiller:
I found a new boundary for SDN_ADM0 as noted in the issue, but this ADM0 does not include a small piece of territory in the Northeast of Sudan that the previous file did. This chunk of land turns out to be the [Halaib Triangle](https://en.wikipedia.org/wiki/Halaib_Triangle), which is disputed between Egypt and Sudan. I figured we shouldn't assign it to Sudan, and the data I got is from 2021. Let me know if you want me to change anything, and I would be happy to note this all in a wiki page as well!

sgmurphy216 (changes requested):
Thanks a bunch for your research Dorian! For all boundaries, our policy is to offer a representation of the country as they represent themselves; therefore, if Sudan claims the Halaib Triangle, we should offer a file that includes it. We'll treat EGY as claiming that territory in their boundaries as well.

dorianvmiller:
Awesome, thanks for your help Sean! I found a new source for the boundary that includes the disputed territory. I also reuploaded the ADM1 boundary because the one I had before didn't include that territory either but now it does (and hopefully includes more than just one of the states this time)!

sgmurphy216 (approved):
Looks great!! Thanks for the correction Dorian! Closes #1407

Issue #1410, opened 2021-06-02
gB OPEN SDN_ADM1


sgmurphy216:
SDN_ADM1 fails build script (geometric). Please find a new source and upload.

PR #1438, opened 2021-06-03
SDN_ADM1

dorianvmiller:
Found new source and file for SDN_ADM1

 #1410


dorianvmiller:
Uploaded only a selection of the shapefile the first time, second upload has the full ADM1 boundaries.

sgmurphy216 (changes requested):
Hey Dorian! I think this second submission also only contains one ADM1 boundary, mind reuploading?

sgmurphy216 (approved):
Thanks for the fix Dorian! Closes #1410

Issue #1439, opened 2021-06-04
[BOUNDARY ERRATA] Encoding errors in meta.txt files


DanRunfola:
**Describe the Error**
The following source metadata files have encoding errors - all information in the metadata files need to be in UTF-8.  We need to fix these before we push 4.0; should be as simple as downloading and fixing the encoded issues in the meta.txt.  

**Screenshots**
![image](https://user-images.githubusercontent.com/7882645/120833607-f89b6380-c52f-11eb-844c-07205d4eec09.png)

**Additional context**
In the gbOpen Dataset, I identified the following boundaries as being impacted:

Romania ADM0 - DONE - https://github.com/wmgeolab/geoBoundaries/pull/1444
Romania ADM1 - DONE - https://github.com/wmgeolab/geoBoundaries/pull/1446
Romania ADM2 is also incorrect (it is UTF-8, but accents have been stripped)

Mali ADM0 - DONE - https://github.com/wmgeolab/geoBoundaries/pull/1447
Mali ADM1 - DONE - https://github.com/wmgeolab/geoBoundaries/pull/1452
Mali ADM2 - DONE - https://github.com/wmgeolab/geoBoundaries/pull/1453
Mali ADM3 - DONE - https://github.com/wmgeolab/geoBoundaries/pull/1454

PR #1447, opened 2021-06-04
MLI ADM0 gbOpen Metadata Correction

kpupkiewicz:
Text in metadata is in UTF-8 & includes license screencap.

#1439


sgmurphy216 (approved):
Looks good! Thank you! Progresses #1439

Issue #1440, opened 2021-06-04
[BOUNDARY ERRATA] Saint Lucia ADM1 Metadata


DanRunfola:
**Describe the Error**
The source is described as both OSM and a digitization.  A bit unclear. 

**Screenshots**
![image](https://user-images.githubusercontent.com/7882645/120834011-75c6d880-c530-11eb-9518-5bfb0caa8510.png)

**Additional context**
Need to correct for clarity.

PR #1449, opened 2021-06-05
LCA_ADM1 Issue #1440

dfornatora:
 https://github.com/wmgeolab/geoBoundaries/issues/1440


sgmurphy216 (approved):
Great work! Thanks! Closes #1440

Issue #1378, opened 2021-05-28
[BOUNDARY ERRATA] MDA-ADM1-3_0_0-B20


leeberryman:
**Describe the Error**

MDA-ADM1-3_0_0-B20 & MDA-ADM1-3_0_0-B34 appear to be geometry mistakes and not legitimate multipart polygons

PR #1450, opened 2021-06-06
MDA_ADM1

dorianvmiller:
Found new source for MDA_ADM1 to resolve boundary errors.

 #1378


dorianvmiller:
Found a new source for MDA_ADM1 that does not have the polygon errors in the previous file - the district named "Cruileni" was one of the two that was identified in the issue because it has two polygons, but this is correct (check out the district map on [this page](https://en.wikipedia.org/wiki/Criuleni_District)). Let me know if this works or if anything should be changed!

sgmurphy216 (approved):
Thanks Dorian! Looks great! Closes #1378

PR #1452, opened 2021-06-07
MLI ADM1 gbOpen Metadata Correction

kpupkiewicz:
Text in metadata is in UTF-8.

#1439


sgmurphy216 (approved):
Looks good! Progresses #1439

Issue #590, opened 2020-10-30
Add ISO Code Sheet to gb Website


sylviashea1:
Add sheet of ISO codes with country and continent labels. Make it available on GitHub and on geoBoundaries website; especially useful for API calls. 
Sample:
Continent  | Country | ISO Code

PR #1455, opened 2021-06-07
Issue #590 ISO Code Sheet

hadleyday:
 https://github.com/wmgeolab/geoBoundaries/issues/590


sgmurphy216 (approved):
Thanks Hadley- this'll be a great update. Closes #590

DanRunfola:
Hey @hadleyday - we need to integrate this with our existing codesheet, which is over at https://github.com/wmgeolab/geoBoundaryBot/blob/main/dta/iso_3166_1_alpha_3.csv

Basically, need to make any changes you made to the attached xlsx in that file, making sure we don't have duplicates etc.

Issue #1462, opened 2021-06-10
[BOUNDARY ERRATA] BGD ADM3 - Metadata


DanRunfola:
**Describe the Error**
The metadata for BGD ADM3 gbOpen is incorrect - it is labeled as ADM2, when I (believe) it should be ADM3. Will need to confirm.

PR #1474, opened 2021-06-10
BGD_ADM3 gbOpen Meta

nmspage:
 #1462 

## Anything Unusual?

 This is @dfornatora 's PR


sgmurphy216 (changes requested):
This file looks good- please include Bangladesh Bureau of Statistics as Source 2 and we should be good to go!

sgmurphy216 (approved):
Looks good! Thanks @dfornatora and @nmspage - closes #1462

PR #1496, opened 2021-06-10
gB Authoritative CIV_ADM0-2

dorianvmiller:
 #1466


sgmurphy216 (approved):
Looks great! Thanks! Closes #1466

PR #1499, opened 2021-06-11
gB Authoritative CAN_ADM0-2

dorianvmiller:
 #1465


sgmurphy216 (approved):
Looks good! Thanks Dorian! Closes #1465

PR #1501, opened 2021-06-11
gB Authoritative HTI_ADM0-2

dorianvmiller:
 #1470


sgmurphy216 (approved):
Looks good! Closes #1470

PR #1508, opened 2021-06-11
TGO_ADM2 gbAuthoritative

rsritchey:


sgmurphy216 (changes requested):
Hey! Looks like you accidentally submitted the BNDA source file from UN SALB instead of the processed data. Also, it looks like you may have submitted the data in a nested folder- please compress the data without putting it into a folder first, then submit. Thank you!!

Issue #1484, opened 2021-06-10
gB Authoritative ZMB


sgmurphy216:
Expansion of gB Authoritative; data at https://www.unsalb.org/data/zmb

*Metadata and license at https://www.unsalb.org/sites/default/files/wysiwyg_uploads/docs_uploads/SALB_DataSpecifications_v2.0_20210601.pdf
List license type as:
UN SALB Data License

DanRunfola:
Closed by #1510 (ADM1, 2) and #1484 (ADM0)

PR #1510, opened 2021-06-14
ZMB_ADM2 Issue #1484

dfornatora:
 https://github.com/wmgeolab/geoBoundaries/issues/1484


nmspage (approved):
Great work!

Issue #1487, opened 2021-06-10
[METADATA UPDATE] gB OPEN KNA_ADM1


sgmurphy216:
Currently have on geoboundaries.org ; please update metadata to reflect new licensing under OCHA

PR #1514, opened 2021-06-14
KNA_ADM1 Issue #1487

dfornatora:
 https://github.com/wmgeolab/geoBoundaries/issues/1487


nmspage (approved):
Thanks!

Issue #1478, opened 2021-06-10
gB Authoritative NZL


sgmurphy216:
Expansion of gB Authoritative; data at https://www.unsalb.org/data/nzl

*Metadata and license at https://www.unsalb.org/sites/default/files/wysiwyg_uploads/docs_uploads/SALB_DataSpecifications_v2.0_20210601.pdf
List license type as:
UN SALB Data License

carolinamonrivera:
@sgmurphy216 License needs to be added to the list for all checks to pass.

DanRunfola:
https://github.com/wmgeolab/geoBoundaries/pull/1533

PR #1519, opened 2021-06-17
NZL_ADM2 gbAuthoritative

carolinamonrivera:
0-2NZL Issue #1478


sgmurphy216 (changes requested):
Hey! Please submit only NZL2 and change your metadata 'license notes' to be on only one line, as this confuses the bot. Otherwise, files look good. Thanks!

PR #1522, opened 2021-06-21
SEN_ADM2 gbAuthorative

carolinamonrivera:
issue #1481

https://github.com/wmgeolab/geoBoundaries/issues/1481


sgmurphy216 (changes requested):
Hey Caro! Your data looks great, but I think you may have included a different meta.txt than the one you intended- please resubmit with the SEN_ADM2 meta and we should be golden! Thanks!

sgmurphy216 (changes requested):
Hey! A few little metadata changes- please change Source 2 and link to source data fields to reflect SEN and we should be good to go. Thanks!

sgmurphy216 (dismissed):
Looks great Caro! Thank you! Closes #1481

sgmurphy216 (approved):

PR #1524, opened 2021-06-22
PHL_ADM0-2 Issue #1479

dfornatora:
 https://github.com/wmgeolab/geoBoundaries/issues/1479


sgmurphy216 (approved):
Looks great- thanks a bunch! Closes #1479

PR #1528, opened 2021-06-23
URY_ADM2 issue #59

rsritchey:


sgmurphy216 (changes requested):
Hey Renee! Could you find a source that is not GADM? We try to refrain from using their data due to previous licensing issues- thanks!

PR #1533, opened 2021-06-27
NZL_ADM1 gbAuthoritative

carolinamonrivera:
Issue #1478


sgmurphy216 (changes requested):
Hey Caro! File looks great, just gotta add the ISO_Code field and we should be set! Thanks!

sgmurphy216 (changes requested):
Hey! Few changes here- please change the fields to 'Name' and 'ISO_Code' and add 'Level', and in meta.txt, please change sources to:

Source 1: From SALB Data, United Nations
Source 2: Source Government (Federal Office of Topology, Afghanistan, etc)

Once that's set we should be golden! Thanks!

sgmurphy216 (approved):
Looks good! Thank you Caro! Closes #1478

Issue #1288, opened 2021-04-29
shapeID = ECU-ADM1-1590546715-B23


leeberryman:
### What is the expected behavior?
Here is one minor error in the 3_0_0 GeoBoundary data. In ADM1, Ecuador, for shapeID = ECU-ADM1-1590546715-B23, shapeName = SANTO DOMINGO DE LOS TSACHILAS the shapeISO should be EC-SD. That would match the ISO code. Currently there is an underscore is used instead of a dash (EC_SD). This is the only case of an underscore being used in the whole world, which looks like a mistake.

### Any possible solutions?

[ ] I'll work on a PR for this!

DanRunfola:
This should now be corrected in GB 4.0.

PR #1539, opened 2021-06-29
gB OPEN ECU_ADM1 #1288

sgmurphy216:
Addressing an attributional inconsistency @leebarryman pointed out- thanks Lee!

#1288


nmspage (commented):
Please fix the capitalization in the ISO_Code field and it will be good to approve :)

nmspage (changes requested):
Please fix the capitalization in the ISO_Code field and it will be good to approve :)

Issue #1537, opened 2021-06-28
[BOUNDARY ERRATA] gB OPEN VNM_ADM2


maxmalynowsky:
ADM2 layer has assigned separate ID's to each island of an ADM2 (each single part geometry), rather than a single ID to the entire multipart of an area.

Expected ADM2 units: 673
Actual ADM2 units: 1,561

<img width="699" alt="Screen Shot 2021-06-29 at 00 04 12" src="https://user-images.githubusercontent.com/4202166/123709547-8dfaf080-d86d-11eb-92c9-34766bf10183.png">

sgmurphy216:
Thanks for bringing this to our attention! @MichaelR-geoLab found a new data source with far less features, but some repeating names persist- canonically, some are simply named the same. Updated data source created by VNM gov't; will be addressed with a wiki page as well.

DanRunfola:
Closed by https://github.com/wmgeolab/geoBoundaries/pull/1540 , also see the brief note at https://github.com/wmgeolab/geoBoundaries/wiki/VNM

PR #1540, opened 2021-06-29
VNM_ADM2 Issue #1537

MichaelR-geoLab:
added fixed Vietnam ADM2


sgmurphy216 (approved):
Looks good after research into repeating district names- it appears this is canonical, and the data creator appears to be VNM gov't. Approving with wiki page- thanks a bunch Michael! Closes #1537

PR #1558, opened 2021-07-06
CHL_ADM0 gbAuthoritative

hadleyday:


sgmurphy216 (changes requested):
Looks great Hadley! Would you mind changing the decompressed filenames to just CHL_ADM0 from CHL_ADM0AUTH? Thanks a bunch!

Issue #1564, opened 2021-07-09
gB OPEN SLB_ADM3


sgmurphy216:
Current file is direct permission- in an attempt to replace direct permission files with open licensed, please find new source.

sgmurphy216:
Update: Cannot use HDX source for this, appears a sourcedata projection error

11helenab:
Source has been changed from HDX to direct permission, apparently through email, is it saved anywhere? The projection of the layer has also been changed to WGS 84. I think this issue has already been fixed and can be closed.

PR #1572, opened 2021-07-09
gB OPEN SLB_ADM3 #1564

rsritchey:


sgmurphy216 (commented):
This has some weird errors- the geometry itself looks fine, projection is correct, meta and all that is fine, but it appears the polygons are placed outside the bounds of a normal EPSG:4326 map. Gonna keep looking at this for a fix

Issue #1563, opened 2021-07-09
gB OPEN SLB_ADM2


sgmurphy216:
Current file is direct permission- in an attempt to replace direct permission files with open licensed, please find new source.

PR #1576, opened 2021-07-11
SLB_ADM2 gbOpen

rmbiddle:
 #1563


sgmurphy216 (approved):
Great find Russ. Closes #1576

Issue #1561, opened 2021-07-09
gB OPEN FJI_ADM3


sgmurphy216:
Current file is direct permission- in an attempt to replace direct permission files with open licensed, please find new source.

PR #1580, opened 2021-07-12
FJI ADM3 gbOpen

kpupkiewicz:

#1561


sgmurphy216 (dismissed):
Looks great! Thank you! Closes #1580

DanRunfola:
Oooo - this has a proper topological error we can't auto-correct.

Could you load this file into mapshaper.org and see if there is a quick/obvious way to fix the errors?

kpupkiewicz:
I used the 'clean' function on mapshaper which appeared to fix the issue, but I think the bot failed it because when mapshaper exported the file, it changed around parts of the attribute table. Going to submit the same boundary with a fixed attribute table and hopefully it will work :E

sgmurphy216 (approved):
Looks great! Thanks! Closes #1561

PR #1581, opened 2021-07-12
AUS_ADM2 gbOpen

hadleyday:
AUS_ADM2

 https://github.com/wmgeolab/geoBoundaries/issues/1169


sgmurphy216 (approved):
Looks good! Thanks Hadley! Closes #1169

Issue #1565, opened 2021-07-09
gB OPEN SLB_ADM4


sgmurphy216:
Current file is direct permission- in an attempt to replace direct permission files with open licensed, please find new source.

kpupkiewicz:
Closing because SLB only officially goes down to ADM3. See wiki page

PR #1582, opened 2021-07-13
SLB_ADM4 gbOpen

kpupkiewicz:

#1565


MichaelR-geoLab (changes requested):
I took a look at your file, it seems there are a lot of repeat names, and I do not think that's correct. Could you check that out for me?

kpupkiewicz:
Good catch!! Found this quote from their official 2009 census, which afaik is the origination for pretty much all SLB data. https://www.statistics.gov.sb/component/advlisting/?view=download&format=raw&fileId=414 on page 5.

> "This consists of a hierarchical system comprising three different levels:   o The national level; Solomon Islands   o The provincial level: nine provinces plus Honiara City Council   o The ward level: 183 wards divided over the ten province level entities

> For field operational purposes the whole country was divided into 1,344 enumeration areas (EAs) which were defined within the ward boundaries, thus constituting a fourth level. Although the division into EAs does not have any official administrative status and differs per census, it may serve as sampling frame for specialized surveys during inter-censual periods. "

So it seems like there wouldn't be official names for these EAs, but it also seems like this boundary maybe doesn't belong in gB after all 😓 If we want to keep it, I think the best option would just to suffix the names (Samasodu1, Samasodu2, etc.)??

kpupkiewicz:
Closing because SLB only officially goes down to ADM3. See wiki page

Issue #1525, opened 2021-06-23
gB OPEN URY_ADM2


sgmurphy216:
Missing layer for build; either find new source or add metadata from geoboundaries.org file

MichaelR-geoLab:
Any source for Uruguay ADM2 is either incomplete or not an open source. I cannot even find a complete map to use for digitization.

dorianvmiller:
Looks like URY ADM3 are municipalities, and they don't cover the whole country - some parts are not divided into them. I found [this map](https://upload.wikimedia.org/wikipedia/commons/5/5a/Mapa_de_Uruguay_%28subdivisiones%29.svg) online and can use it to digitize if we want this layer.

DanRunfola:
Yes, let's move forward with a digitization for now unless we can find an
alternative.  We'll also need a wiki entry just to note that it isn't
contiguous.

rsritchey:
@dorianvmiller I can digitize this if you want

PR #1583, opened 2021-07-13
URY_ADM2

dorianvmiller:
Digitized layer for URY municipalities - I used several images for this, so let me know if the license.txt needs to be reformatted in any way.

 #1525


sgmurphy216 (approved):
Looks great Dorian- thanks for doing this. Please add a wiki page explaining the non-continuous nature of the ADM2 level with a source or two and we'll be golden! Closes #1525

Issue #1585, opened 2021-07-14
[BOUNDARY ERRATA] Zambia LIcense Clarification


DanRunfola:
From our GRID3 partners, some updates we need to make to the Zambia case:

GRID3 Zambia boundaries, the primary source is the Office of the Surveyor General. The boundaries represent legislative changes of 2017 however the shapefile was updated in 2020. Also, the license is CC by 4.0 not international.

PR #1586, opened 2021-07-14
ZMB_ADM2 gbOpen Metadata Update

kpupkiewicz:

#1585


sgmurphy216 (approved):
Thanks for the update Kristina- looks great. Closes #1585

PR #1591, opened 2021-07-15
PLW_ADM2 gbOpen

kpupkiewicz:

I found a good resource for the south pacific while working on other boundaries, so expect a few island nations coming soon


sgmurphy216 (approved):
File looks great Kristina! Thanks for finding that source, looks to be very handy.

Issue #1566, opened 2021-07-09
gB OPEN NER_ADM3


sgmurphy216:
Current file is direct permission- in an attempt to replace direct permission files with open licensed, please find new source.

PR #1597, opened 2021-07-20
NER_ADM3 gbOpen

rmbiddle:
 #1566


sgmurphy216 (changes requested):
This submission looks good but contains two unneeded pngs; niger_com.png and niger_com1.png . Please resubmit without these pngs and we should be good to go! Thanks!

sgmurphy216 (approved):
Looks great Russ- thanks a bunch! Closes #1566

PR #1605, opened 2021-07-21
LCA_ADM2 gbOpen

kpupkiewicz:


sgmurphy216 (changes requested):
Appears humdata downloaded this data from GADM- let's try to find a different source for LCA2

sgmurphy216 (approved):
Looks great Kristina! Thank you for finding this.

Issue #1536, opened 2021-06-28
[BOUNDARY ERRATA] gB OPEN PRT_ADM3


maxmalynowsky:
ADM3 units from different ADM2's appear to be merged together into a multipolygon that all share the same ID. When performing a join by location with the ADM2, this results in strange artifacts.

ADM3 units sharing the same ID:
<img width="1174" alt="Screen Shot 2021-06-28 at 22 48 43" src="https://user-images.githubusercontent.com/4202166/123708898-77a06500-d86c-11eb-91ea-ad2f03234708.png">

Join by location between ADM3 and ADM2 layers
<img width="1175" alt="Screen Shot 2021-06-28 at 22 48 54" src="https://user-images.githubusercontent.com/4202166/123708934-84bd5400-d86c-11eb-9f2b-edf6efcbfd4e.png">

Actual ADM2 layer, with location of ADM3 units sharing same ID
<img width="1175" alt="Screen Shot 2021-06-28 at 22 49 05" src="https://user-images.githubusercontent.com/4202166/123708968-90a91600-d86c-11eb-8cf9-28983aaabc11.png">

sgmurphy216:
Thanks for spotting this Max- @hadleyday 's working on it now!

PR #1606, opened 2021-07-21
PRT_ADM3 gbOpen

hadleyday:
PRT_ADM3

 https://github.com/wmgeolab/geoBoundaries/issues/1536

## Anything Unusual?

 This pull request separates the 6 parishes named Pinhiero into their own individual polygons, rather than one large multipolygon


sgmurphy216 (changes requested):
Thank you so much for working on this- bit of a complex issue, but I appreciate the extra effort it took to verify the subunit names/correct the multipolygon error. If you wouldn't mind renaming shapeName and shapeType fields to our standard Name and Level , I think we're good to go!

DanRunfola:
Moved to https://github.com/wmgeolab/geoBoundaries/pull/1683

PR #1607, opened 2021-07-22
gB OPEN SLB_ADM3 is#1564

sgmurphy216:
Original file had a projection error; reuploading with corrected projection


sgmurphy216:
Odd- appear to be getting the same geometric bounding error after reprojecting to WGS84 in Arc from source data (https://data.humdata.org/dataset/solomon-islands-administrative-level-0-1-2-and-3-boundary-polygons). 
Visually, we don't have the same issue as last time though: 
image

@DanRunfola thoughts?

Issue #1322, opened 2021-05-06
gB Authoritative Slovenia ADM0


slfuhrig:
Possible source if dissolve water management areas shaepfile: http://nfp-si.eionet.europa.eu/Dokumenti/GIS/splosno

DanRunfola:
No longer relevant with 4.0 and redefinition of Authoritative to SALB.

PR #1609, opened 2021-07-23
SVN_ADM0 gbAuth

rmbiddle:
 #1322


sgmurphy216 (changes requested):
File appears to be missing geometry- please reupload! Thanks!

sgmurphy216:
Looks like this PR is accidentally submitting to both gB Auth and gB OPEN -- please resubmit PR into only gB Auth

sgmurphy216 (changes requested):
Looks like these files are actually missing geometry- please reupload with the full file and we'll be good!

Issue #1608, opened 2021-07-23
gbOpen VEN ADM1 - faulty character encoding


karimbahgat:
**Describe the Error**
gbOpen Venezuela ADM1 appears to have a faulty character encoding. E.g. look at the topojson link below, and click on the different admins on the map, shows some wonky names. 

https://github.com/wmgeolab/geoBoundaries/blob/main/releaseData/gbOpen/VEN/ADM1/geoBoundaries-VEN-ADM1.topojson

PR #1610, opened 2021-07-23
VEN_ADM1 Issue #1608

MichaelR-geoLab:
Got a new source with good encoding.


sgmurphy216 (approved):
Looks great Michael- thanks a bunch for the fix! Closes #1608

PR #1645, opened 2021-07-28
NZL_ADM2 gbAuthoritative

carolinamonrivera:
issue #1478


sgmurphy216 (changes requested):
Looks good Caro- would you mind changing source 2 to the government office the file came from on UN SALB? This should be the same as the field in the NZL_ADM1 meta.txt file. Thanks!

sgmurphy216 (approved):
Thanks Caro! Closes #1478

Issue #1489, opened 2021-06-10
gB OPEN OMN_ADM2


sgmurphy216:
Please locate and upload- we have this in non-open license on geoboundaries.org

PR #1647, opened 2021-07-29
OMN_ADM2 gbOpen

carolinamonrivera:
gbOpen Issue #1489

https://github.com/wmgeolab/geoBoundaries/issues/1489


sgmurphy216 (changes requested):
Hey Caro- could you elaborate on how you got direct permission for this file? We're trying to remove direct perms files from all gB releases, but the license png looks to be compatible with gB Open. The tweet appears to be a reference to an ODbL type license.

DanRunfola:
@sgmurphy216 @carolinamonrivera 
I just did a dig into this - it looks like the license is likely acceptable, but something we'll want to replace in 5.0.  Going to go ahead and accept for now.

Issue #1653, opened 2021-07-29
gB OPEN TON_ADM2


sgmurphy216:
Tonga is separated into 'divisions' and 'districts'; 5 divisions and ~23 districts. This district level would be useful to find for our 4.0 release, as previous data was incorrect.

PR #1655, opened 2021-07-29
TON_ADM2 Issue #1653

dfornatora:
 https://github.com/wmgeolab/geoBoundaries/issues/1653


sgmurphy216 (changes requested):
File appears to be somewhat incomplete; are there other data sources we could use?

image

sgmurphy216 (approved):

sgmurphy216:
Closes #1653

Issue #1661, opened 2021-07-30
gB OPEN GMT_ADM2


sgmurphy216:
Need a new source for GMT2

PR #1663, opened 2021-08-01
GTM_ADM2 Issue #1661

dfornatora:
 https://github.com/wmgeolab/geoBoundaries/issues/1661


sgmurphy216 (approved):
Great work as always Dominic- closes #1661

Issue #1621, opened 2021-07-27
[BOUNDARY ERRATA] BHS ADM2


DanRunfola:
Need to determine if BHS ADM2 exists.  If not, we need a wiki page for it.  If so, we need it :).

PR #1673, opened 2021-08-03
Issue #1621 BHS_ADM2

hadleyday:
BHS ADM2

 https://github.com/wmgeolab/geoBoundaries/issues/1621


sgmurphy216 (changes requested):
Files look great but I think you may have forgotten to include license.png- add that and we're good to go!

sgmurphy216 (approved):
Looks great Hadley! Thank you! Closes #1621

Issue #1670, opened 2021-08-02
gB OPEN BHS_ADM1


sgmurphy216:
BHS_ADM1 missing boundary for the island of Nassau- this is an issue, as Nassau is not only a significant portion of landmass, but also the capital (oops!). Need a repair or new data source for 4.0

PR #1684, opened 2021-08-06
Issue #1670 BHS_ADM1

hadleyday:
 https://github.com/wmgeolab/geoBoundaries/issues/1670


sgmurphy216 (changes requested):
Hey Hadley! Looks like this submission only contains one geometry for BHS_ADM1 and the 'level' field contains ADM2- think you may have submitted the wrong file. Mind reuploading? Thanks!

sgmurphy216 (changes requested):
Hey Hadley! Your second submission is perfect, but I think this PR is attaching both the original erroneous file and the correct final one- mind submitting a new PR with only the second file? Thank you so much!

Issue #1688, opened 2021-08-12
[BOUNDARY ERRATA] Hole in MDA_ADM0


MichaelR-geoLab:
**Describe the Error**
A clear and concise description of what the error is - please provide sources as appropriate.

Hole in MDA_ADM0

**Screenshots**
If applicable, add screenshots or images to help explain your problem.

**Additional context**

DanRunfola:
This was closed by https://github.com/wmgeolab/geoBoundaries/pull/1689

PR #1689, opened 2021-08-12
Issue #1688 MDA_ADM0 fix

MichaelR-geoLab:


sgmurphy216 (changes requested):
The fix looks great- we just need to change source 1 to geoBoundaries and create a wiki page explaining the alterations to the original geometry (patched hole)- then we should be good to go!

DanRunfola:
@sgmurphy216 I force-accepted this PR so I could get it into the 4.0 build that is happening now.  I'm going to create an issue to get your comments addressed shortly after the 4.0 release.

Issue #1755, opened 2021-09-11
gB Open ALB ADM2


lydiatroup:
Need a new file for ALB ADM2

(assigned to mikayla)

PR #1805, opened 2021-09-15
ALB_ADM2

mlwilliams01:
 #1755


lydiatroup (commented):
This is a really good first pull request!! We actually can't use this boundary though because it's source is GADM (they're one of our competitors) There are only 3 small things that would need to be changed.

- your license png is titled "license.png.png" so you'd want to just title it "license" and it'll fill in the .png part
- ^ same is true for the text file, it's titled "meta.txt.txt" so yo'd want to just title it "text" and it'll fill in the .txt
- for boundaries coming from this site (humdata) the source is actually what's listed unser "source" in the metadata on the site, so for this boundary it would be gadm and unhcr. which is why we unfortunately can't use it.

lydiatroup:
I'm going to go ahead and close this pull request since we can't use boundaries from gadm, but otherwise it looked really good!

Issue #1756, opened 2021-09-11
gB Open ALB ADM3


lydiatroup:
Need a new file for ALB ADM3

(assigned to Mikayla)

ejott:
Going to remove my assignment from this because I'm not getting anywhere with it, but to leave a record of progress, [this source](https://geoportal.asig.gov.al/geonetwork/srv/eng/catalog.search#/metadata/6508921c-f12b-44de-957b-2077ef625cdd) _might_ be public use ("no restrictions for public use") but there's no way to download it from that site, and sources elsewhere (i.e. INSTAT's ArcGIS Hub) have no license information attached so it's a bit shaky.

PR #1806, opened 2021-09-15
ALB_ADM3

mlwilliams01:
 #1756


lydiatroup:
going to go ahead and close this one too. you'll need to find a different source for this and alb adm2!

Issue #1724, opened 2021-09-07
gB Authoritative LKA


lydiatroup:
Need UN SLAB data for LKA (ALL ADM Levels EXCEPT ADM2 and ADM4)

lydiatroup:
need adm 3 for this one as well (if it exists)

hadleyday:
No data on ADM 3, only can upload levels 0-2

DanRunfola:
https://github.com/wmgeolab/geoBoundaries/pull/1808 does ADM1

DanRunfola:
https://github.com/wmgeolab/geoBoundaries/pull/1807 Does ADM0

PR #1807, opened 2021-09-16
Issue #1724 LKA_ADM0

hadleyday:
LKA ADM 0

 https://github.com/wmgeolab/geoBoundaries/issues/1724


lydiatroup (approved):
Looks great! Thanks for the quick work!

PR #1808, opened 2021-09-16
Issue #1724 LKA_ADM1

hadleyday:
LKA ADM1

 https://github.com/wmgeolab/geoBoundaries/issues/1724


lydiatroup (changes requested):
meta and license look good, but your shapefile is empty! try exporting again & reuploading

lydiatroup (approved):
looks great! closes #1724

Issue #1748, opened 2021-09-11
gB Authoritative MEX ADM1


lydiatroup:
Need the salb data for mexico adm1

PR #1810, opened 2021-09-17
MEX_ADM1

taraamcl:
#1748


lydiatroup (changes requested):
Looks mostly good! There's only two things you need to change, the source and links in the meta should be changed to align with the salb info i sent you! And the ISO codes in the shapefile should be more specific (bc it's an adm1) you can find them [here](https://www.iso.org/obp/ui/#iso:code:3166:MX)

lydiatroup:
new pr, closing this one

Issue #1717, opened 2021-09-07
gB Authoritative CYP


lydiatroup:
Need UN SLAB data for CYP (ALL ADM Levels)

PR #1812, opened 2021-09-17
CYP_ADM0 and CYP_ADM1

taraamcl:
#1717 (This copy should be correct-- meta files had the wrong name)


lydiatroup:
There's a couple things you need to change, the source and links in the meta should be changed to align with the salb info i sent you! And the ISO codes in the ADM1 shapefile should be more specific (bc it's an adm1), and the adm0 attribute table is empty.

lydiatroup (changes requested):
There's a couple things you need to change, the source and links in the meta should be changed to align with the salb info i sent you! And the ISO codes in the ADM1 shapefile should be more specific (bc it's an adm1), and the adm0 attribute table is empty.

lydiatroup:
new pr submitted, closing this one

Issue #1799, opened 2021-09-12
gB OPEN GEO_ADM1


sgmurphy216:
Layer has a bad license for 5.0. Replace with acceptable license type (no virals, no ODbL)

PR #1815, opened 2021-09-17
GEO_ADM1-2 gbOpen

rmbiddle:
 #1799


sgmurphy216 (changes requested):
Hey Russ! Please source your data in meta.txt from the 'Contributor' or 'Source' from HDX, not from HDX directly (gB --> HDX integration makes this problematic). Thanks!

Issue #1751, opened 2021-09-11
gB Open PAK ADM2


lydiatroup:
need a new file for PAK ADM2 - no more osm

assigned to amanda

leeberryman:
@lydiatroup I have an adm2 and adm1 that the geometry alligns with the adm3 file for PAK that my team contributed a few months back.  Do you want us to submit those too?

leeberryman:
@sgmurphy216 I see all the issues being created to deprecate ODbL license in 5.0.  Some of the data my team created we released with this license, but if this isn't an acceptable license for GeoBoundaries uses I'd like to change it.  Let me know so we can adjust.  Thanks.

sgmurphy216:
> @sgmurphy216 I see all the issues being created to deprecate ODbL license in 5.0.  Some of the data my team created we released with this license, but if this isn't an acceptable license for GeoBoundaries uses I'd like to change it.  Let me know so we can adjust.  Thanks.

Working up acceptable license list for 5; will share ASAP. Also, any additional PAK contributions would be great! The full list of our boundaries is now available at www.geoBoundaries.org under the 'View Data' tab- any contributions we're missing are more than welcome.

Thanks as always Lee.

PR #1818, opened 2021-09-18
Issue #1751 PAK_ADM2

amanda-reed:
PAK_ADM2
[PAK_ADM2.zip](https://github.com/wmgeolab/geoBoundaries/files/7190048/PAK_ADM2.zip)

#1751 

## Anything Unusual?

<!--- You can also include anything else we need to know while assessing this data. -
->


lydiatroup:
this wasn't uploaded properly, but it'll be easy to fix! I'm sending you a message to sort it out.

Issue #1723, opened 2021-09-07
gB Authoritative SEN


lydiatroup:
Need UN SLAB data for SEN (ALL ADM Levels EXCEPT ADM3)

lydiatroup:
we already had this country in gb authoritative

PR #1819, opened 2021-09-19
Issue #1723 SEN_ADM0

hadleyday:
SEN ADM0

 https://github.com/wmgeolab/geoBoundaries/issues/1723


lydiatroup:
we actually already had this country in authoritative

PR #1821, opened 2021-09-20
CYP_ADM1, CYP_ADM0

taraamcl:
#1717


sgmurphy216 (changes requested):
Hey Tara- looks like this file did not contain geometries. Try deselecting all elements in Arc and resubmitting. Thanks!

PR #1828, opened 2021-09-21
CYP Authoritative

charliealtman:
You're welcome
 CYP ADM0&1

 #1717


sgmurphy216 (changes requested):
Hey Charlie!! This is an interesting one. First off- not that interesting- it appears your data source has a small slice out of it in CYP ADM0. We'll have to talk about how to proceed with this as we don't edit the geometries of SALB files. Since this is an internal issue, may not be too much of a problem.
Secondly, and much more interestingly, our bot only accepts files from 1990 onwards. Since this data is validated as of 1960, we'll have to make an exception. Will reach out with some guidance soon.

Good work!

lydiatroup:
closed bc new pr made

Issue #1783, opened 2021-09-12
gB OPEN CZE_ADM0-1


sgmurphy216:
Layer has a bad license for 5.0. Replace with acceptable license type (no virals, no ODbL)

DanRunfola:
https://github.com/wmgeolab/geoBoundaries/pull/1834 closes this for ADM1

PR #1829, opened 2021-09-21
CZE_ADM0 #1783

rsritchey:


sgmurphy216 (approved):
Looks great as always Renee! Keep up the good work! Progresses #1783

PR #1830, opened 2021-09-21
CYP ADM 0,1 gb Auth

charliealtman:
YUH
 CYP ADM0&1

 #1717


lydiatroup (changes requested):
bot is failing because the year of the data is old (1960), but that's a problem for dan! both your files' metadata has source 2 and link to source data for romania. please fix and reupload and it'll be perfect

lydiatroup:
new one uploaded, closing this

PR #1831, opened 2021-09-22
gbOpen SLB_ADM1

carolinamonrivera:
ISSUE #1798

https://github.com/wmgeolab/geoBoundaries/issues/1798


lydiatroup (changes requested):
same is true for this one, we can't use humdata anymore, so you will need to find a new source!

carolinamonrivera:
Finding a new source

PR #1833, opened 2021-09-22
SLB_ADM0

carolinamonrivera:
gbOpen Issue 1798

https://github.com/wmgeolab/geoBoundaries/issues/1798


lydiatroup (changes requested):
We can no longer use humdata for our boundaries. you'll need to find a different source for this boundary!

carolinamonrivera:
finding a new source

PR #1861, opened 2021-09-24
CYP Authoritative

charliealtman:
YUH
 CYP ADM0&1

 #1717


lydiatroup (approved):
mission SUCCESS! I am approving this with the note that it's only failing because of the old year (1960), but that is for dan to fix!

DanRunfola:
Initiating a re-run on the metachecks to resolve this.  I updated the bot to accept older years (up to 1950) in response to this.

PR #1864, opened 2021-09-26
Issue #1748 MEX_ADM1

hadleyday:
MEX_ADM1

 https://github.com/wmgeolab/geoBoundaries/issues/1748


lydiatroup (approved):
looks great, thanks for the quick work hadley!

Issue #1823, opened 2021-09-21
gB Open DEU ADM0


lydiatroup:
Need a new file for DEU ADM0 (no humdata, share-alike, or odbl)

PR #1869, opened 2021-09-28
Add files via upload

ssheminway:
DEU_ADM0

 #1823

## Anything Unusual?

 GeoReferenced with 16 control points and the Projective and Spline Transformations


sgmurphy216 (approved):
Pushing this through - one ring self-intersect, we'll see if the simplification algorithm fixes it. If not- simple fix, we'll run it through a 1% correction in mapshaper and see if that corrects. Great first submission though! Closes #1823

Issue #1722, opened 2021-09-07
gB Authoritative ROU


lydiatroup:
Need UN SLAB data for ROU (ALL ADM Levels)

DanRunfola:
https://github.com/wmgeolab/geoBoundaries/pull/1817 does this for ADM1
Still need ADM2

PR #1871, opened 2021-09-29
ROU_ADM0 Issue #1722

dfornatora:
 https://github.com/wmgeolab/geoBoundaries/issues/1722


lydiatroup (approved):
looks great, thanks for the quick work! closes #1722

Issue #1820, opened 2021-09-20
NPL ADM2 field name not as expected from GeoBoundaries


leeberryman:
'DISTRICTS' is the field name in the shapefiles, geojson, topojson etc instead of the expected value of shapeName

DanRunfola:
We did a manual direct edit in this case, after failing to find better licensed sources for nigh on a year and half :)

Should be in the DB sometime in the coming week.

PR #1878, opened 2021-10-07
NPL_ADM2 Issue #1820

dfornatora:
Correcting attribute table field names

 /issues/1820


lydiatroup (changes requested):
This file looks good, but we can't use any licenses with share-alike clauses anymore, so you'll need to find a new file

lydiatroup:
bad license, closing pr

Issue #1732, opened 2021-09-08
[BOUNDARY ERRATA] SLV ADM2


DanRunfola:
SLV ADM2 is sourced from GADM, via humdata, and needs to be replaced with a more open license.

SidonieH:
Found public domain image on wikipedia, will work on digitizing

PR #1883, opened 2021-10-08
#1732 ADM2

SidonieH:
 #1732

## Anything Unusual?

 There are some empty polygons


lydiatroup (changes requested):
It looks like you submitted two files in two different locations instead of one. Please upload the correct file (with the meta.txt file) to sourcedata/gbopen in a new pull request and @ me in the comments i will close this one! (it's just easier to make a new pr instead of deleting the files in this one and trying to fix it here in the cases of multiple files uploaded)

Issue #1854, opened 2021-09-23
gB OPEN KOR ADM1


lydiatroup:
need a new file for KOR ADM1 (no humdata, share-alike, or odbl)

PR #1886, opened 2021-10-10
KOR_ADM1 gbOpen

carolinamonrivera:
Issue: 1854

https://github.com/wmgeolab/geoBoundaries/issues/1854


sgmurphy216 (changes requested):
Submission looks great- one small fix, please change the name of the .zip file to KOR_ADM1 as opposed to KOR-ADM1 . Little change, but consistency is the name of the game :)

PR #1907, opened 2021-10-24
ROU_ADM0 gbOpen

carolinamonrivera:
Issue #1786

 https://github.com/wmgeolab/geoBoundaries/issues/1786


carolinamonrivera:
@sgmurphy216 the license listed on the site is not specifically listed on the gb license list -- let me know if this is a valid pr. im pretty sure its not a share-alike but the only "unported" license on the list includes the sharealike feature. On https://www.statsilk.com/license it is specifically listed as "Creative Commons Attribution 3.0 Unported License" not "Creative Commons Attribution-ShareAlike 3.0 Unported" -- I'm not sure if these are the exact same license but I know that if it is the share-alike this is not a good source. Please let me know, I'm confused loll.

lydiatroup:
following up on this @carolinamonrivera  the error in metadata is coming from your license. It should be "Creative Commons Attribution-ShareAlike 3.0 Unported" with no parentheses

Issue #1847, opened 2021-09-23
gB OPEN DNK_ADM1


lydiatroup:
need a new file for DNK adm1 (no humdata, share-alike, or odbl)

PR #1912, opened 2021-10-29
DNK ADM1 gBOpen - Issue #1847

MichaelR-geoLab:
Digitized Denmark ADM1


sgmurphy216 (changes requested):
Hey Michael- 

This looks really good, just need you to calculate the LEVEL field and add ISO_Code from https://www.iso.org/obp/ui/#search - then we should all be set. Thanks for digitizing this, I know the tiny islands can get annoying. Great work.

sgmurphy216 (approved):
Thanks for the fixes Michael- closes #1847

Issue #1792, opened 2021-09-12
gB OPEN NRU_ADM1


sgmurphy216:
Layer has a bad license for 5.0. Replace with acceptable license type (no virals, no ODbL)

PR #1915, opened 2021-11-02
NRU_ADM1 #1792

amanda-reed:
 #1792


lydiatroup (changes requested):
accidently requested changes on this and am not sure how to update that, but id try submitting a pr for this file again and tagging me in it and we can see if it runs? I know weve tried this a couple times but maybe itll work now

lydiatroup (approved):
Looks great, thanks for the quick fix! closes #1792

Issue #1853, opened 2021-09-23
gB OPEN DOM_ADM2


lydiatroup:
need a new file for DOM ADM2 (no humdata, share-alike, or odbl)

PR #1916, opened 2021-11-05
gB OPEN DOM_ADM2 #1853

ssheminway:
DOM_ADM2

 #1853


lydiatroup (approved):
looks perfect! closes #1853

Issue #1840, opened 2021-09-22
gB Open HUN ADM1


lydiatroup:
need a new file for HUN adm1 (no humdata, share-alike, or odbl)

PR #1932, opened 2021-11-12
HUN_ADM1 gbOpen

rmbiddle:
 #1840


sgmurphy216 (changes requested):
Hey Russ- all looks good here, please just resubmit the zip file without the hungary_counties.png portion and we should be good to go.

lydiatroup:
@sgmurphy216

Issue #1779, opened 2021-09-12
gB OPEN LBN_ADM0


sgmurphy216:
Layer has a bad license for 5.0. Replace with acceptable license type (no virals, no ODbL)

lydiatroup:
made a new issue for

PR #1933, opened 2021-11-12
LBN_ADM0 gbOpen

rmbiddle:
 #1779


sgmurphy216 (changes requested):
Hey Russ- looks like this is just an error in the license type field of the metadata file. Refer to https://github.com/wmgeolab/geoBoundaryBot/blob/main/dta/gbLicenses.csv to get the exact wording- file looks fine though, so we're all good there.

lydiatroup:
@sgmurphy216

Issue #1859, opened 2021-09-23
gB OPEN SWE_ADM1


lydiatroup:
need a new file for SWE ADM1 (no humdata, share-alike, or odbl)

SidonieH:
I combined individual open source counties. However, two of the counties I can't find any open source image to digitize. Can I just "fill in the blank" from our ADM0 file? Since there are only two left it is pretty easy to figure out.

PR #1943, opened 2021-11-16
#1859

SidonieH:
 #1859


lydiatroup (approved):
looks great, thanks Sidonie! closes #1859

Issue #1841, opened 2021-09-22
gB Open HUN ADM0


lydiatroup:
need a new file for HUN adm0 (no humdata, share-alike, or odbl)

PR #1945, opened 2021-11-16
#1841

SidonieH:
 #1841


sgmurphy216 (changes requested):
Hey Sidonie - looks like you just have to rename your metadata txt file from 'Meta.txt' to 'meta.txt' , that should resolve us!

lydiatroup:
@sgmurphy216

PR #1948, opened 2021-11-16
#1755 ALB_ADM2

mlwilliams01:
 #1755


sgmurphy216 (changes requested):
Hey! Think we just need to correct the 'canonical boundary name' field in the metadata text file to 'County'. Thank you!

sgmurphy216 (approved):

Issue #1824, opened 2021-09-21
gB Open DEU ADM1


lydiatroup:
Need a new file for DEU ADM1 (no humdata, share-alike, or odbl)

PR #1950, opened 2021-11-16
#1824 DEU_ADM1

mlwilliams01:
 #1824


sgmurphy216 (changes requested):
Hey! Please add the iso 3166-2 codes (subunit level, should be like DEU-something from iso.org's country lookup) in the ISO_Codes field, then we should be good to go!

lydiatroup:
closing bc new pr submitted

Issue #1891, opened 2021-10-11
Add district code (AGS) to Germany ADM2 data


edonnachie:
German administrative districts are identified using a 5-digit code (AGS: https://en.wikipedia.org/wiki/Community_Identification_Number#Germany). This is important to link with other district-level data because the district names are not always consistent.

The AGS seems to be present in your source data, but not in your final package.

**Describe the solution you'd like**
Please add AGS (5 digits) as a attribute to the spatial data.

**Describe alternatives you've considered**
The only alternative is to use a different data source (e.g. GADM.org) or try to match by inconsistent names.

PR #1953, opened 2021-11-19
DEU_ADM2 gbOpen

carolinamonrivera:
issue 1891

https://github.com/wmgeolab/geoBoundaries/issues/1891


carolinamonrivera:
Added a new file for the district codes that Germany uses -- believe this is the reason for the failed pr. @sgmurphy216 @DanRunfola

lydiatroup (changes requested):
one super easy change and this file will be perfect! License was spelled wrong in the metadata license type, so it was failing. please change the license to "Data license Germany - Attribution - Version 2.0" and the bot should pass it!

lydiatroup (approved):
looks great!

Issue #1753, opened 2021-09-11
gB Open ALB ADM0


lydiatroup:
We need a new file for ALB ADM0

(no more osm)

Assigned to selwyn

PR #1958, opened 2021-11-19
#1753 ALB_ADM0 gbOpen

ssheminway:
ALB_ADM0

 #1753


lydiatroup (changes requested):
reopening this from ages ago, but just needs a little attribute table cleaning to be accepted! please make sure theres only 3 fields: Name, ISO_Code, Level titled as such and filled out!

lydiatroup (changes requested):
Screen Shot 2022-04-29 at 1 49 46 PM
this is what your attribute table currently looks like, please make it Name (Albania), ISO_Code (ALB), Level (ADM0) 
and then it'll be perfect!

lydiatroup (approved):
approving this! closes #1753

Issue #1781, opened 2021-09-12
gB OPEN USA_ADM0-2


sgmurphy216:
Layer has a bad license for 5.0. Replace with acceptable license type (no virals, no ODbL)

mmyrte:
I'm unsure if this merits its own ticket, since the license issue could be resolved by switching to another data source - 
The current USA data are not of terrestrial boundaries, but include territorial waters; e.g. admin 2 in southern Florida:

<img width="1287" alt="image" src="https://user-images.githubusercontent.com/24587121/139816847-3fc988eb-970c-4be3-b4b9-7358041983a0.png">

mmyrte:
I got around to checking out the source indicated in the metadata. The [county level data (zip of shp)](https://www2.census.gov/geo/tiger/GENZ2018/shp/cb_2018_us_county_500k.zip), for instance, _do_ follow the coastlines. Please enlighten me if I fundamentally misunderstood the gbOpen data product, but to me, it seems that someone must have manually applied a buffer to the original data. CGAZ actually _does_ follow the coastline.

Moreover given that the data source is the US government, it should be public domain. I could not find an explicit license for the [TIGER products](https://www.census.gov/programs-surveys/geography/guidance/tiger-data-products-guide.html), but [data.gov](https://resources.data.gov/open-licenses/) has a nice overview. They refer to <http://www.usa.gov/publicdomain/label/1.0/> as being a default to refer to. Should I try my hand at making a PR, or is someone else much further along already?

sgmurphy216:
> I got around to checking out the source indicated in the metadata. The [county level data (zip of shp)](https://www2.census.gov/geo/tiger/GENZ2018/shp/cb_2018_us_county_500k.zip), for instance, _do_ follow the coastlines. Please enlighten me if I fundamentally misunderstood the gbOpen data product, but to me, it seems that someone must have manually applied a buffer to the original data. CGAZ actually _does_ follow the coastline.
> 
> 
> 
> Moreover given that the data source is the US government, it should be public domain. I could not find an explicit license for the [TIGER products](https://www.census.gov/programs-surveys/geography/guidance/tiger-data-products-guide.html), but [data.gov](https://resources.data.gov/open-licenses/) has a nice overview. They refer to <http://www.usa.gov/publicdomain/label/1.0/> as being a default to refer to. Should I try my hand at making a PR, or is someone else much further along already?

[...]

PR #1959, opened 2021-11-20
USA_ADM0-2 gbOpen

carolinamonrivera:
issue 1781

 https://github.com/wmgeolab/geoBoundaries/issues/1781


lydiatroup (approved):
look perfect as always! thank you for the submission!

PR #1967, opened 2021-12-02
gbOpen ROU_ADM0

carolinamonrivera:
ISSUE 1786 Fixing meta

 https://github.com/wmgeolab/geoBoundaries/issues/1786


lydiatroup (changes requested):
the 3.0 unported license is actually the same as the normal cc by 3.0, so please change your license to "Creative Commons Attribution 3.0 License" and the bot should pass it!

lydiatroup (approved):
looks great!

PR #1968, opened 2021-12-02
gbOpen BHS_ADM0

carolinamonrivera:
ISSUE 1796

https://github.com/wmgeolab/geoBoundaries/issues/1796


lydiatroup (approved):
looks good! thanks for the submission!

Issue #1754, opened 2021-09-11
gB Open ALB ADM1


lydiatroup:
Need a new file for ALB ADM1

(assigned to selwyn)

PR #1970, opened 2021-12-02
#1754 ALB_ADM1

mlwilliams01:
 #1754


lydiatroup (changes requested):
still a weird binary issue occuring, try deleting the spaces after "License Notes:" in your meta and resubmit! and to resubmit, click on your patch at the top of this pr and navigate to gbopen and upload there!

lydiatroup (changes requested):
one more thing to change. The iso codes in your ISO_code field should be the codes listed in the 3166-2 code column on this site: https://www.iso.org/obp/ui/#iso:code:3166:AL
for adm1 levels we use these more specific codes as these codes are specific to each individual region!

lydiatroup (approved):
looks good! Thank you for all your hard work this semester, happy graduation!!!

Issue #1843, opened 2021-09-23
gB Open VNM ADM1


lydiatroup:
need a new file for NVM adm1 (no humdata, share-alike, or odbl)

PR #1971, opened 2021-12-03
gB Open VNM ADM1 #1843

ssheminway:
 #1843


lydiatroup (changes requested):
one small change, the ISO code field should contain the iso codes mentioned in this [site](https://www.iso.org/obp/ui/#iso:code:3166:VN) (the 3166-2 codes)

lydiatroup (approved):
looks great! thanks for making those changes so quickly!

Issue #1827, opened 2021-09-21
gB Open FRA ADM1


lydiatroup:
Need a new file for FRA adm1 (no humdata, share-alike or odbl licenses)

hadleyday:
Possible leads (need to validate on another computer because I currently don't have space on this one): 
https://www.data.gouv.fr/fr/datasets/contours-geographiques-des-nouvelles-regions-metropole/
https://www.data.gouv.fr/fr/datasets/referentiel-geographique-francais-communes-unites-urbaines-aires-urbaines-departements-academies-regions-1/

PR #1973, opened 2021-12-05
Issue #1827 FRA_ADM1

hadleyday:
FRA ADM 1

 https://github.com/wmgeolab/geoBoundaries/issues/1827


lydiatroup (changes requested):
two changes needed:

1. your iso code field should contain the iso codes from the "3166-2" column in this page: https://www.iso.org/obp/ui/#iso:code:3166:FR
2. your name column is still titled name in france (i'm guessing thats what it says???), please make a new column that's titled "Name"

PR #1976, opened 2021-12-06
Issue #1827 FRA_ADM1

hadleyday:
FRA_ADM1

 https://github.com/wmgeolab/geoBoundaries/pull/1973


lydiatroup (changes requested):
that name column is still titled wrong. You'll want to create a new 'Name' field and calculate it to be = to your french name field, just changing the name of the column in arc won't work

PR #1979, opened 2021-12-11
FRA_ADM1

hadleyday:
FRA ADM1

 https://github.com/wmgeolab/geoBoundaries/issues/1827


lydiatroup (approved):
problem SOLVED looks good!

Issue #2003, opened 2022-01-27
gB Authoritative CMR ADM 0-3


lydiatroup:
date on all UN SALB (authoritative) meta files need to be updated to 2021-02-22.

download existing files from code -> source data -> gb Authoritative

PR #2075, opened 2022-02-01
CMR_ADM0 #2003

ssheminway:
CMR_ADM0

 #2003
 metadata fix


lydiatroup (changes requested):
theres actually no file submitted here, please upload one!

lydiatroup (approved):
looks perfect! progresses #2003

PR #2076, opened 2022-02-01
CMR_ADM1 metadata fix #2003

### LINKED ISSUE #2003: gB Authoritative CMR ADM 0-3

ssheminway:
 #2003


lydiatroup (changes requested):
please update the dates in the metadata to reflect dan's new request (dd-mm-yyyy to dd-mm-yyyy) where the first date is the date on the site (in the chunk that has date to last update) and the second date is the date currently in your metadata

lydiatroup (changes requested):
you'll want to change the meta dates to 2000-01-01 to 2021-02-22 but in the proper dd-mm-yyyy format

lydiatroup (changes requested):
meta dates should be in  dd-mm-yyyy to dd-mm-yyyy format

lydiatroup (changes requested):
your attribute table is missing an ISO_code field! Please add it using the 3166-2 codes from this site: https://www.iso.org/obp/ui/#iso:code:3166:CM
otherwise looks good!

lydiatroup (approved):
looks perfect!

PR #2077, opened 2022-02-01
CMR_ADM2 metadata fix #2003

### LINKED ISSUE #2003: gB Authoritative CMR ADM 0-3

ssheminway:
#2003


lydiatroup (changes requested):
same comment as on the ADM1 layer, please update the dates in the metadata to reflect dan's new request (dd-mm-yyyy to dd-mm-yyyy) where the first date is the date on the site (in the chunk that has date to last update) and the second date is the date currently in your metadata

lydiatroup (changes requested):
you'll want to change the meta dates to 2000-01-01 to 2021-02-22 but in the proper dd-mm-yyyy format

lydiatroup (changes requested):
dates should be in  dd-mm-yyyy to dd-mm-yyyy format

lydiatroup (approved):
this looks good! progresses #2003

Issue #2005, opened 2022-01-27
gB Authoritative BDI ADM 0-2


lydiatroup:
UN Salb layers for Burundi need to be added to gB Authoritative.

PR #2082, opened 2022-02-02
BDI_ADM2

taraamcl:
#2005


lydiatroup (changes requested):
this uploaded weird and honestly i'm not sure why, so make a new zip file when you fix the changes below please
- name field should be titled "Name", super picky but important for standardization
- we actually don't need an ISO_Code field for this file bc it's an ADM2 so please delete that field
- license info should be updated to the following: UN SALB Data License , https//www.unsalb.org/sites/default/files/wysiwyg_uploads/docs_uploads/TermsOfUseSALB2021.pdf
- year should be updated to the date listed on the site "to" the date listed in the attribute table (dd-mm-yyyy format for both dates). if there isn't a date in the original attribute table leave a comment on this pr or the issue

lydiatroup:
closed bc more recent one

Issue #2006, opened 2022-01-27
gB Authoritative MOZ ALL ADM levels


lydiatroup:
date on all UN SALB (authoritative) meta files for all MOZ adm levels need to be updated to the date in 

1) download UN Salb data from un salb site, record date present in the attribute table
2) download existing gB files from code -> source data -> gb Authoritative
3) update metadata date to reflect date found in step 1
4) reupload file

lydiatroup:
no date on the attribute table

PR #2084, opened 2022-02-04
gB Authoritative MOZ ALL ADM levels

dorianvmiller:
Metadata update for all MOZ ADM levels

 #2006


lydiatroup:
@dorianvmiller

Issue #2008, opened 2022-01-27
gb Authoritative ALL layers Metadata update


lydiatroup:
date on **all** UN SALB (authoritative) meta files for **ALL** adm levels need to be updated to the date in

1)download UN Salb data from un salb site, record date present in the attribute table
2)download existing gB files from code -> source data -> gb Authoritative
3)update metadata date to reflect date found in step 1
4)reupload file

this requires going through ALL of the UN salb layers (all gb auth) and updating the dates. this is an URGENT issue to fix

SidonieH:
Costa Rica (CRI) has no date in SALB @lydiatroup

SidonieH:
Haiti (HTI) has no date in SALB @lydiatroup

SidonieH:
- ISL (Iceland) has a earlier date in the attribute table (2016) than the from date on the website (2017)
- ROU (Romania) no date in SALB
- TGO no date in SALB
- ZMB (Zambia) no dat ein SALB

PR #2100, opened 2022-02-08
#2008 gB Administrative LSO 0-2

SidonieH:
 #2008


sgmurphy216 (changes requested):
Hey! This looks good- could you add ISO_Code to LSO_ADM1 and resubmit? Thank you!

SidonieH:
On further inspection of LSO ADM 1 - there appear to be doubled districts in the attributte table (two features with the same name in different locations) and missing some districts compared to the list on https://www.iso.org/obp/ui/#iso:code:3166:LS

lydiatroup:
@SidonieH this is one of the circumstances where we just have to go with what we have because the UN salb data isn't data we're meant to be editing the actual shapes of or anything. And one fix needs to be made for your LSO adm 1 file (make the ISO codes be the more specific ones not just LSO so the ls-d ones on the site you have linked above), otherwise everything is good.

lydiatroup:
lfs uploaded them & closing this

PR #2102, opened 2022-02-08
#2008 gB Administrative NZL 0-2

SidonieH:
 #2008


sgmurphy216 (changes requested):
Hey- some funky fieldnames in the NZL_ADM1 file here. ISO3CD should be ISO_Code and correctly filled for each subdivision - currently each reflect NZL, Also, ADM1NM --> 'Name'

sgmurphy216 (approved):
Looks good! Thanks Sidonie! #2008

Issue #2043, opened 2022-01-30
gB Open FJI ADM2


lydiatroup:
some ADM2s use underscores instead of hyphens in names, fix this

PR #2105, opened 2022-02-08
#2043 FJI_ADM2

jingjli1:
underscores changed to hyphens in ADM2 names.


sgmurphy216 (changes requested):
Hey! The files in here are perfect, looks like you tried to submit into the parent directory before /gbOpen/ - no worries at all, the easiest fix is probably to close out this issue and resubmit into /sourceData/gbOpen/ as the bot will get confused when the first submission on this request is just /sourcedata/FJI_ADM2 . Thanks!

PR #2107, opened 2022-02-09
BDI_ADM0

taraamcl:
#2082 ADM_0


sgmurphy216 (changes requested):
Hey Tara! Few changes and we should be all set:

- Meta text file has to be called 'meta.txt' for the bot to recognize it- yours just has a capital M, so the bot gets confused
- For the 'License: ' field in the meta file, the bot will only recognize 'UN SALB Data License' if memory serves
- Date range should be input as dd-mm-yyyy to dd-mm-yyyy , with dashes as opposed to slashes

Apart from those minor issues, good submission! Thanks for the work.

lydiatroup:
new pr posted

Issue #2016, opened 2022-01-28
gB Open AFG ADM2


lydiatroup:
name column needs to be retitled to "name". Additionally, if working in arc uncheck visibility on the "shapelength" and "object id" columns

PR #2121, opened 2022-02-14
#2016 AFG_ADM2

jingjli1:
Edits to field names and visibility.


sgmurphy216 (changes requested):
Hey Jing! Same general feedback, we'll need to double check that 'open' is an acceptable license delineation. However, just a quick note, we only require the ISO_Code field in ADM0-1, so the field is not needed for this file. Once that's changed and we make sure on the license type, should be good to go! Thanks!

sgmurphy216 (approved):

DanRunfola:
@sgmurphy216 I'm ready to accep this, but what issue number does it close?

sgmurphy216:
> @sgmurphy216 I'm ready to accep this, but what issue number does it close?

#2016

DanRunfola:
@jingjli1 thank you for the AFG contributions - hugely appreciated, and a geometry we've struggled to find.

Issue #1974, opened 2021-12-05
gB OPEN BHS ADM1


lydiatroup:
Layer has a bad license for 5.0. Replace with acceptable license type (no virals, no ODbL)

PR #2127, opened 2022-02-15
gbOpen BHS_ADM1 #1974

ssheminway:
 #1974


lydiatroup:
dd-mm-yyyy to dd-mm-yyyy and we should be perfect

lydiatroup (approved):
looks good! closes #1974

Issue #2049, opened 2022-01-30
gB Open IND ADM0


lydiatroup:
layer needs to be replaced

DanRunfola:
Closed by https://github.com/wmgeolab/geoBoundaries/pull/2138

PR #2128, opened 2022-02-15
IND ADM0

shauntmathew:
IND_ADM0
 Digitized from wikimedia commons

 #2049


sgmurphy216 (changes requested):
Already fixed- resubmit with singlepart to multipart poly dissolution

PR #2135, opened 2022-02-18
gbAuthoritative NER_ADM0

dorianvmiller:
New file for NER_ADM0 with updated metadata

 #2093


sgmurphy216 (approved):
Looks great! Progresses #2093

PR #2136, opened 2022-02-18
gbAuthoritative NER_ADM1

dorianvmiller:
New file for NER_ADM1 with updated metadata

 #2093


sgmurphy216 (approved):
Looks good! Progresses #2093

PR #2142, opened 2022-02-22
gbOpen AFG_ADM1 issue #2014

rsritchey:


lydiatroup (changes requested):
Hi Renee! There a couple issues with this file that need to be fixed before we can accept it;
- your license image should have the link visible in it
- source 1 in the meta should be geoBoundaries and source 2 would actually be the USGS (United States Geological Survey)
- the shapefile is in the wrong coordinate system, please project it to "wgs 1984" (under geographic coordinate systems if you're working in arc, EPSG4326-WGS84 in Q)
- there are some holes and overlaps in this file, make sure you're using the trace feature when digitizing to prevent these errors. in the image below the pink regions have an invalid topology that the bot won't accept
Screen Shot 2022-02-23 at 6 13 20 PM

rsritchey:
@lydiatroup That is so weird, normally these mistakes never happen so I think I am just going to have to redo this one.

DanRunfola:
@rsritchey poke!

lydiatroup (changes requested):
This file needs to be projected into the right coordinate system (EPSG 4326 - WGS84) and the word document in the zipfile needs to be deleted

PR #2140, opened 2022-02-21
NGA_ADMX

taraamcl:
#2106, all levels

PR #2146, opened 2022-02-23
NGA_ADM1

taraamcl:
(PR #2140)


DanRunfola:
UN SALB updated NGA, which Sidonie has integrated the new versions of.  So these are now deprecated.

Issue #2158, opened 2022-02-25
gB Administrative CRI ADM 0-2,  SEN ADM 0-2


SidonieH:
NEW FILE needed for gb Authoritative CRI and SEN
dates needed fixing

lydiatroup:
still needs cri adm2 upload

PR #2160, opened 2022-02-25
gB Authoritative CRI ADM1

SidonieH:
 #2158


lydiatroup (changes requested):
please change the iso codes in the attribute table to the more specific codes found on online browsing at this site https://www.iso.org/iso-3166-country-codes.html

lydiatroup (approved):
looks great, thanks!

Issue #2040, opened 2022-01-30
gB Open ECU ADM2


lydiatroup:
names contain encoding errors for special characters "Ã°±", please fix and reupload

PR #2161, opened 2022-02-25
ECU_ADM2

amanda-reed:
 #2040


lydiatroup (approved):
looks great! closes #2040

Issue #2094, opened 2022-02-07
gB Authoritative ZWE ADM 0-2


lydiatroup:
NEW FILE needed for gb Authoritative ZWE.

- data should be downloaded from UN Salb, website.
- dates should be formatted as: "dd-mm-yyyy to dd-mm-yyyy" (with first date being the one on the site and second being the one in the attribute table)
- need to upload all adm levels: 0-2. to do so, will need to use the dissolve tool to dissolve using the adm1 field (for adm1) or no field (for adm0).
- when uploading all three files, upload in separate zipfiles and upload as usual, but upload three zipfiles instead of just one.

lydiatroup:
license info: 
UN SALB Data License
https//www.unsalb.org/sites/default/files/wysiwyg_uploads/docs_uploads/TermsOfUseSALB2021.pdf

lydiatroup:
closing this issue bc prs were merged!

PR #2172, opened 2022-03-04
ZWE_ADM1 gbAuthoritative

dorianvmiller:
 #2094


lydiatroup (changes requested):
your attribute table is missing an iso code field, please add one with the 3166-2 iso codes form this site: https://www.iso.org/obp/ui/#iso:code:3166:ZW

looks good otherwise!

Issue #2181, opened 2022-03-06
gB Auth TGO ADM0-2 metafix


lydiatroup:
please upload these as separate PRs.
can use our current file ( in release data) , just need to update the date to "29-06-2017 to 25-10-2021"

PR #2188, opened 2022-03-06
TGO_ADM0 Issue#2181

dfornatora:
correcting date in meta. Issue #2181

/issues/2181


lydiatroup (approved):
looks good! thanks dominic

Issue #2114, opened 2022-02-11
gB Open UGA ADM3


lydiatroup:
getf.org has requested we add these boundaries

PR #2198, opened 2022-03-10
UGA_ADM3

shauntmathew:
UGA ADM3

 #2114 

## Anything Unusual?
 There are two licenses since this boundary was digitized via two pngs.


sgmurphy216 (approved):
Great work Shaun, thanks for getting this in. Closes #2114

lydiatroup (changes requested):
this file was uploaded to the wrong location! if you look underfiles changed the path is just the zipfile and not sourcedata/gbopen/uga_adm3.zip. This is why the bot didn't run. Please resubmit under a new pr this file uploaded to the correct location!

PR #2199, opened 2022-03-22
#2008 gB Administrative LSO 1

SidonieH:
 Fixing LSO  1 #2008


lydiatroup (approved):
looks good!
because this fix came out of a multifile pr, you'll have to submit another pr with 0 and 2 for lso (together or individually, they should both work for now) so we can get all the lso files merged.

PR #2201, opened 2022-03-26
issue #1841 HUN_ADM0

shauntmathew:
HUN ADM0

 #1841


lydiatroup (changes requested):
only one small thing to fix! There's a space at the end of your shapefile names and your zipfile. What you'll want to do to fix this is to reexport from arc making sure there is not a space at the end of the name of the file and then create a new zipfile and a **new** pr for this file

lydiatroup:
new pr created

PR #2202, opened 2022-03-26
#2114 UGA_ADM3

shauntmathew:
UGA_ADM3

 #2114


lydiatroup (changes requested):
thank you for fixing the upload location on this so quick! One small thing i'd change, since the source for both images is Wikipedia Commons i'd make source 2 be geoBoundaries since source 1 of wikipedia commons covers both images. other than, that looks good!

lydiatroup (approved):
jk this looks perfect! closes #2114

Issue #2055, opened 2022-01-30
gB Open IRL ADM2


lydiatroup:
current file is wrong, needs to be replaced

PR #2207, opened 2022-03-29
#2055 gbOPEN IRL_ADM2

SidonieH:
 #2055


lydiatroup (changes requested):
a couple notes on this file:
1) we have a wiki on ireland rightnow regarding adm2s in ireland that currently states ireland adm2s dont exist, but your source contradicts that as does the [wikipedia page](https://en.wikipedia.org/wiki/Local_electoral_area). Use your best judgement about whether or not this adm level is administrative and update the wiki if you think it is!
2) I think your Canonical boundary type name has a typo in the meta; assuming you mean "Local Electoral Areas"
3) we don't need an iso code field for adm2s, so please delete that field in the attribute table.

lydiatroup (approved):
looks good! closes #2055

Issue #2225, opened 2022-04-02
gbOpen LCA ADM1


shauntmathew:
**Describe the Error**
License Improvement Request - Current ADM1 layer has same issue as ADM0 that it was sourced from Open Street Map and is not Public Domain

**Screenshots**
If applicable, add screenshots or images to help explain your problem.

**Additional context**

PR #2226, opened 2022-04-02
#2225 LCA ADM1

shauntmathew:
Pull request to an issue I made for a license improvement request for Saint Lucia ADM1, similar to Saint Lucia ADM0

 LCA ADM1

 #2225 
 Usable license for 5.0


shauntmathew:
I needed to use two images to digitize this because my initial digitization was marginally outdated. I used the same metadata and upload format as UGA ADM3 which also had 2 licenses

lydiatroup (approved):
looks good! thanks shaun! closes #2225

Issue #2220, opened 2022-03-31
NOR_ADM1 Attribute table update


lydiatroup:
this iso codes in the attribute table all appear as "null." they should be the iso codes as found on this [site](https://www.iso.org/iso-3166-country-codes.html) -> open browsing platform, using the codes listed in the 3166-2 codes column

**use the file found in [this pr](https://github.com/wmgeolab/geoBoundaries/pull/1989) and update the iso codes in the attribute table and reupload to that existing pr** (ask if you need a refresher on how to do that)

SidonieH:
How do I download the data? Do I need to run the code?

PR #2234, opened 2022-04-07
NOR_ADM1 #2220

SidonieH:
 #2220 @lydiatroup


lydiatroup (changes requested):
this looks good! going to confirm with dan that this is the right way to upload for this before changing to an approve.

lydiatroup (approved):
dan approved! #2220

Issue #2239, opened 2022-04-12
VAT ADM0 License Improvement request


shauntmathew:
**Describe the Error**
A clear and concise description of what the error is - please provide sources as appropriate.

**Screenshots**
If applicable, add screenshots or images to help explain your problem.

**Additional context**

PR #2241, opened 2022-04-15
#2239 VAT_ADM0

shauntmathew:
VAT ADM0

 #2239


lydiatroup (approved):
looks good! funky shape and all, happy easter closes #2239

Issue #2110, opened 2022-02-10
[BOUNDARY ERRATA] ZMB ADM1 Update


DanRunfola:
New Zambia ADM1 is now available:
https://zambia-nsdi-mlnr.hub.arcgis.com/datasets/6fe71ff483ab491782c8524ec3d1a3b8_0/about

PR #2247, opened 2022-04-22
ZMB_ADM1

dfornatora:
ZMB ADM1 for Issue #2110

 /issues/2110


lydiatroup (approved):
looks great! thanks for the hard work closes #2110

Issue #2030, opened 2022-01-30
gB Open CIV ADM3


lydiatroup:
names contain encoding errors showing "Ã©" In place of accented characters, fix this

PR #2248, opened 2022-04-25
CIV ADM3 #2030

jingjli1:
Corrected encoding error in names


lydiatroup (changes requested):
Hi Jing! 
We actually can't use anything from humdata. If you can trace it back to the source humdata is drawing from we can use that data if it has an appropriate license.

PR #2255, opened 2022-04-26
gbOpen ABW #2253

leeberryman:


lydiatroup (changes requested):
Hi Lee!
Thank you for your contribution, the file itself looks perfect and we appreciate you contributing to fill in the gaps in our database. However, there is a slight issue with the date. The 'date representative of' field for the source you used would be the 1914 - 1917 range, which is too old to be accepted into our database. If you wanted to digitize a new file for this here are some links to newer images you could use! [link1](https://commons.wikimedia.org/wiki/File:Aruba_map_ru.svg), [link2](https://commons.wikimedia.org/wiki/File:Aruba_map.png)

leeberryman:
@lydiatroup as I'm looking at these small island nation/territories that we identified that don't currently have any corresponding shape in geoboundaries, I'm finding it difficult to find images or maps without share alike constraints.  I do have the ability to digitize from sentinel-2-l2a imagery since these are all islands this works.  Please see https://scihub.copernicus.eu/twiki/pub/SciHubWebPortal/TermsConditions/Sentinel_Data_Terms_and_Conditions.pdf for the data license for sentinel-2-l2a imagery  can you or @DanRunfola affirm that this license will be acceptable for digitizing the island territories that are missing?

lydiatroup:
@leeberryman This will be a question for @DanRunfola, I'll let him know about the problem.

DanRunfola:
Yes - I think we're safely in the realm of derived works here if Sentinel imagery is used (as the sentinel data isn't in-and-of-itself boundary information).  As Sentine is mostly a public good anyway, I think we're on quite firm ground.
@leeberryman apologies for the delayed response, I am just now returning from a paternity leave!

-Dan

DanRunfola:
Pinging this to keep it open for now

DanRunfola:
Pinging this - we'll be returning to this in the Fall and I don't want it to get marked stale.

lydiatroup (approved):
I'll go ahead and approve this, thank you for the contribution Lee and again, apologies on the delay! closes #2253

Issue #2028, opened 2022-01-28
gB Open CHN ADM1


lydiatroup:
need a new file for chn adm1

PR #2256, opened 2022-04-27
#2028 CHN_ADM1

VictorGedeck:
CHN_ADM1

 https://github.com/wmgeolab/geoBoundaries/issues/2028


lydiatroup (changes requested):
a couple issues with this file, remind me during class and I can talk to you more about it

- license type is not compatible, we can't use any license that says 'share a like' in it and unfortunately this is one of those. this means you'll have to find a new image to digitize. I'd suggest picking an adm0 so you could get it done this week.
- your meta data isn't completely filled out, check out the 'what to upload to github' file in the drive
- your file is showing up as only one feature. This normally means you had something selected when you exported, a super common problem just be sure to deselect everything before exporting
- license image should just be titled 'license' not adm name and level and license

lydiatroup:
closing for wrong license type

PR #2259, opened 2022-04-27
gbOpen BLM_ADM0 #2258

leeberryman:


lydiatroup (changes requested):
Hi Lee! Thank you for the contributions as always!
The ABW file wasn't opening for some reason, so below are comments on the BLM file only
- We actually cannot use any "share-alike" licenses, so the file type is incompatible with our database
- Year representative of refers to the year the image being digitized was created, so I believe in this case it would be 2015, not 2022

Let me know if you have any questions!

Issue #2009, opened 2022-01-27
gB Open JOR ADM0


lydiatroup:
need a new file for JOR ADM0

PR #2260, opened 2022-04-29
JOR_ADM0 #2009

rsritchey:


lydiatroup (changes requested):
i think the bot is down which is making this fail BUT it also has a handful of issues
- we cannot use any share-alike licenses. This has been true since fall 2021
- Canonical Boundary Type Name type field is not filled in
- license.png should be a picture of the spot on the page where it says the license type, not the page on the creative commons site
- also please add a note in the metadata "other notes" section that it was digitized

Issue #2263, opened 2022-04-30
gB Open LBN ADM0


shauntmathew:
**Describe the Error**
A clear and concise description of what the error is - please provide sources as appropriate.

**Screenshots**
If applicable, add screenshots or images to help explain your problem.

**Additional context**

shauntmathew:
Digitized from OSM, needs usable license for 5.0

PR #2265, opened 2022-04-30
#2263 LBN_ADM0

shauntmathew:
LBN_ADM0

 #2263


lydiatroup (changes requested):
This is failing because your "Link to Source" in your meta is missing the colon between "Link to Source" and the url!

Issue #2267, opened 2022-05-01
gB Auth SWZ ADM0


lydiatroup:
please upload this asap! use the document in the drive (linked on the resources doc) for help!

PR #2272, opened 2022-05-02
SWZ_ADM0 Issue #2267

dfornatora:
Issue #2267

 /issues/2267


lydiatroup (changes requested):
there's an error in your attribute table! the level should be ADM0 (i assume) and not ADM1

lydiatroup (approved):
great, thank you for the change! also a note, please hide the shape length and area default fields in the attribute table before future uploads. These are something we'll have to remove all of eventually so might as well do them sooner than later!

Issue #2266, opened 2022-05-01
gB Open SOM_ADM2


lydiatroup:
please download the file [in this PR](https://github.com/wmgeolab/geoBoundaries/pull/2238) and upload it with the following changes to the meta (below). Please ensure both the pr and this issue are linked in the new PR

License: CC-BY Attribution 4 (our standard CC-bY, I thin it's 4 now!)
Source-1: Global Polio Eradication Initiative (GPEI)
Source-2: Frank Salet

PR #2278, opened 2022-05-04
SOM_ADM2 Issue #2266

dfornatora:
For Issue #2266, from pr 2238

 /issues/2266


lydiatroup (approved):
looks great, closes #2266

Issue #2274, opened 2022-05-04
gB Open MYS ADM0


lydiatroup:
Need a new file for this! Current file is an outdated osm layer.
Michael - could you dissolve your adm3 MYS file (thus creating the adm0) and upload it ?

jingjli1:
pr submitted and approved.

PR #2279, opened 2022-05-05
MYS_ADM0 upload Issue #2274

MichaelR-geoLab:
dissolved adm3 layer and included missing Sabah region


lydiatroup (commented):
two small things,
- for digitized files we want source 1 to be geoboundaries and source 2 to be the image's source
- this license is actually not one we can use. Anything with the 'share-alike' we cannot use. because of that I'm just going to let this PR auto close but you do not have to make any changes since the semester is over!

Issue #1784, opened 2021-09-12
gB OPEN TWN_ADM0-1


sgmurphy216:
Layer has a bad license for 5.0. Replace with acceptable license type (no virals, no ODbL)

PR #2282, opened 2022-05-08
#1784 TWN_ADM0

VictorGedeck:


lydiatroup (changes requested):
only one small thing to change!
we don't include the image used to digitize in our files since it can be found at the link in the metadata, so please delete that image and reupload and it should be good to go!

lydiatroup:
closing bc new pr uploaded

Issue #2236, opened 2022-04-08
[BOUNDARY ERRATA] AFG ADM1


sgoodm:
Non-standard schema for AFG ADM1 boundary attributes 

![image](https://user-images.githubusercontent.com/9094862/162497575-a1c820b8-a64d-447e-a8b6-ad7af9b7989e.png)

PR #2285, opened 2022-05-12
Standardized Attribute Table for AFG_ADM1 #2236

ssheminway:
AFG_ADM1

 #2236
 The table has been standardized and a few spelling mistakes corrected

## Anything Unusual?
 I modified the current data and the License isn't one of the typical ones we use - might want to check it


lydiatroup (approved):
looks good! closes #2236

Issue #2061, opened 2022-01-30
gB Open ISR ADM1


lydiatroup:
current file has naming inconsistencies and is license is outdated, please find a new file and reupload

PR #2286, opened 2022-05-12
gB Open ISR ADM1 #2061

ssheminway:
ISR ADM1

 #2061 
 updated license


lydiatroup (approved):
looks good! closes #2061

Issue #2063, opened 2022-01-30
gB Open KEN ADM3


lydiatroup:
Most ADM3 names end with "Ward" verify that this is used colloquially. Other ADM3 names are all caps which should be converted to proper case

PR #2361, opened 2022-09-07
KEN_ADM3 gbOpen

rmbiddle:
 #2063


lydiatroup (changes requested):
this is a great find russ! 3 changes needed
1) license type should be "Creative Commons Attribution 4.0 International (CC BY 4.0)"
2) please make a new field titled 'Name' to replace the field currently titled as ward
3) in the meta change "Wards" to "ward" (should be singular!)

lydiatroup (changes requested):
same problem as the BLM PR - the reupload was in the wrong location. to do a file correction on github you'll click on your patch AND navigate to the source data -> gB Open folder and then upload the file there.

Please make a new PR for this issue with the corrected file uploaded to the correct file path.

Issue #2295, opened 2022-07-26
Auto Update of UN OCHA data


MichaelBrandonFalk:
Hello,

There are several countries missing info or with outdated data in the UN OCHA column, but I am able to find the data on humdata. Would it be possible for Geoboundaries to periodically pull from humdata ( maybe monthly). This would resolve the Pakistan issue whish is now tagged as Urgent. It would also update many other countries including Rwanda (OCHA's layer is really good for Rwanda) , Ethiopia, and many other countries. Thank you.

DanRunfola:
Hi MIchael,

We've been debating the value of mirroring HDX for a little while now, though this is tentatively on the list if we do.  If possible, could you provide a little more detail on your use case?  I.e., why do you use geoBoundaries to access the UN OCHA data, as contrasted to humdata?

-Dan

leeberryman:
@DanRunfola, @MichaelBrandonFalk is one of the stakeholders in consuming a dataset that we have built downstream as a pipeline from geoboundaries, geonames, OSM hamlets and age/sex pop grid values from Worldpop 100m.  At this point we only pull from gBOpen, but all of our stakeholders are non profits and the gBHumanitarian release is suitable for their needs, so @MichaelBrandonFalk has pointed out that in many cases the OCHA data is more recent than gBOpen and in several cases OCHA has newer data than gBHumanitarian.  Our pipeline is already built out to get all the appropriate source and attribution data from the GeoBoundaries API, if their are more development hours needed to make the mirror happen I'm happy to talk through what needs to happen so that we could meet everyone's needs.

DanRunfola:
Got it - and, yes, we would anticipate the OCHA data to be ahead of ours in many cases (with the tradeoff being licensure).

@karimbahgat built a web interface and hooks that automatically generate pull requests for us to process, which I think would be the appropriate way to do this (i.e., a nightly cron --> new HDX file identified --> Pull request generated --> we pull it in after a check).  He can provide more details on the API, but you can also see it in action here: https://www.geoboundaries.org/gbContribute.html

behind the scenes for the hookds: https://github.com/wmgeolab/gbWeb/blob/main/gbContribute.html

DanRunfola:
(and the nice thing about that API is that it's something anyone can ping against - i.e., we don't need to host the chron; any user could).

leeberryman:
Thanks @DanRunfola I'll create a work item for one of our team members to look at this and they might reach out as they have questions.

[...]

PR #2362, opened 2022-09-07
gbHumanitarian 2022-09-07

maxmalynowsky:

https://github.com/wmgeolab/geoBoundaries/issues/2295

## Anything Unusual?

Bulk submission of entire current OCHA Common Operational Datasets


DanRunfola:
I am finally integrating this into the DB (as it was too big to generate a diff, I did a manual checkout and am running checks now).  A huge thank you to @maxmalynowsky for pulling this in.

A few that aren't going to make the cut, just FYSA.  I'm setting this entire system up to do something approaching nightly builds in the very near future, so this won't be a "let's wait for Dan for 3 months" issue ever again (i.e., we should be able to track down the more specific licenses manually in these cases soon).  I also want to take another crack at automating the entire HDX ingest pipe, but just not there yet.

MMR ADM0,1,2,3 - "Other" license
LBN ADM 0,1,2,3 - "Other" license
NER ADM 0,1,2,3 - "Other" license
HND ADM 0,1,2 - "Other" license
TUR ADM 0,1,2 - "Other" license

DanRunfola:
@all-contributors add @maxmalynowsky  to data, code, review, issue

PR #2367, opened 2022-09-08
AUS_ADM0 gbOpen

maxmalynowsky:

Update AUS_ADM0 with 2021 source.


lydiatroup (changes requested):
Hi Max! Overall, this looks great just a couple of small changes needed;

1. For our ADM0s the canonical boundary type name field in the meta text file should be the name the country refers to itself as (so "the Commonwealth of Australia" rather than "Country")
2. The NAME field in the attribute table should just be capitalized as "Name" 

Please make these changes and reupload to this PR, or we can make the changes and reupload for you. Thank you for all your submissions!

lydiatroup (changes requested):
Canonical boundary type name in the meta file still needs to be changed to "Commonwealth of Australia" as that is the full name the country refers to itself as. This change will need to be reflected in the Name field of the attribute table as well.

Please make these changes and reupload, thank you!

maxmalynowsky:
Before I make the change to the boundary type name, what list do you use to get that name? I'll be editing a number of other ADM0 files after this, so it would be good to know how to handle those cases too. From what I'm aware of, this is the ultimate authority for country names: https://unstats.un.org/unsd/geoinfo/ungegn/docs/26th-gegn-docs/wp/wp54_ungegn%20wg%20country%20names%20document%202011.pdf

According to this list, I'd be using "National Official / Formal Name" for AUS to get "the Commonwealth of Australia". However, for countries like Switzerland with multiple official languages, it's not obvious which one to pick, and I'd assume for geoBoundaries the choice would be to use "UN Official / English / Formal name". However under this logic, that would be just "Australia" for AUS.

The above works for countries, but for ISO 3166-1 alpha-3 representing territories, those names aren't included in the above. For those dataset submissions, the list I would pull from would be the following: https://unstats.un.org/unsd/methodology/m49/overview/

lydiatroup:
@maxmalynowsky We had a chance to discuss this internally, and agree that going with the UN Official / English / Formal name would be the way to go going forwards to standardize our naming system! Thank you for making us aware of this document.

lydiatroup (approved):
switching this PR to approved!

PR #2368, opened 2022-09-08
AUS_ADM1 gbOpen

maxmalynowsky:

Update AUS_ADM1 with 2021 source.


lydiatroup (approved):
This file looks perfect! Thank you for the contribution, Max!

Issue #2355, opened 2022-09-05
[BOUNDARY SUBMISSION] gB Open AUS


maxmalynowsky:
Current ADM2 for Australia comes from combining state level datasets, resulting in gaps / overlaps along state boundaries, and is missing names for ADM2, using only codes. A version made available by the government of Australia provides more recent data up to ADM4.

ADM4 based on the 2016 Census:
- Original:  https://datapacks.censusdata.abs.gov.au/datapacks/
- Topologically cleaned: https://data.fieldmaps.io/tmp/aus_adm4.gpkg.zip

This data source is highly standardized and doesn't need much additional processing, however it contains some null geometries representing migratory overseas addresses that I've removed. These boundaries are also used for CIESIN's Gridded Population of the World, which is how I came across this dataset.

maxmalynowsky:
I was just reviewing this, and I saw that a pull request had already been made, but it was for the wrong Shapefile, and I noticed that I had used the wrong data sources as well. Statistical Areas go down to 4 levels, but they are not the same as administrative divisions. This pull request uses Statistical Areas Level 3 as ADM2:

- https://github.com/wmgeolab/geoBoundaries/pull/2243

The correct file to use for ADM2 should be Local Government Areas:

- https://www.abs.gov.au/statistics/standards/australian-statistical-geography-standard-asgs-edition-3/jul2021-jun2026/access-and-downloads/digital-boundary-files/LGA_2022_AUST_GDA2020_SHP.zip

Sources for identifying LGA as ADM2:

- https://demos.mapbox.com/boundaries-explorer/?country=AU&layer=adm2&worldview=US
- https://en.wikipedia.org/wiki/List_of_administrative_divisions_by_country

maxmalynowsky:
Since there's currently an error with the Heroku app submission tool, here are links to cleaned and formatted boundaries ready to be added to [gbOpen](https://github.com/wmgeolab/geoBoundaries/tree/main/sourceData/gbOpen):

- https://data.fieldmaps.io/tmp/AUS_ADM0.zip
- https://data.fieldmaps.io/tmp/AUS_ADM1.zip
- https://data.fieldmaps.io/tmp/AUS_ADM2.zip

PR #2369, opened 2022-09-08
AUS_ADM2 gbOpen

maxmalynowsky:

Update AUS_ADM2 with 2022 source. 
Fixes previous boundaries by changing Statistical Area 3 for Local Goverment Area. 
Remove file from git lfs.


lydiatroup (approved):
This is a beautiful file! Thank you for the submission, Max! clsoes #2355

PR #2370, opened 2022-09-08
CHE_ADM0 gbOpen

maxmalynowsky:

Update CHE_ADM0 with 2022 source.
Removed polygons from Italy, Germany, and Liechtenstein.


lydiatroup (changes requested):
Please change the canonical boundary type name field in the meta data to reflect the name of the country; "The Swiss Confederation"
Additionally, please change the name of the country in the attribute table to reflect that name change.

Thank you!

PR #2373, opened 2022-09-08
CHE_ADM3 gbOpen

maxmalynowsky:

Update CHE_ADM3 with 2022 source.
Removed polygons from Italy, Germany, and Liechtenstein.
Removed from git lfs.


lydiatroup:
Closing this PR as one is being uploaded through LFS now that I have verified the correctness of the file!

PR #2374, opened 2022-09-08
USA_ADM0 gbOpen

maxmalynowsky:

Update USA_ADM0 with 2021 source.
Removed US territories and moved to separate files.


lydiatroup (changes requested):
This file still contains some territories; please remove those. Additionally please change the canonical boundary type name field in the meta file to "The United States of America"

PR #2377, opened 2022-09-08
ASM_ADM0 gbOpen

maxmalynowsky:

Create ASM_ADM0 with 2021 source.


DanRunfola:
Fixed and integrated in https://github.com/wmgeolab/geoBoundaries/pull/2449

maxmalynowsky:
Could I suggest re-opening this since the source from https://github.com/wmgeolab/geoBoundaries/pull/2449 is missing an island, and that ADM0 won't align with the ADM1 from the US census bureau?

PR #2381, opened 2022-09-08
GUM_ADM0 gbOpen

maxmalynowsky:

Create GUM_ADM0 with 2021 source.


lydiatroup (changes requested):
Please change the canonical boundary type name field in the meta file to the name of this ADM (Guam)

PR #2383, opened 2022-09-08
GUM_ADM2 gbOpen

maxmalynowsky:

Create GUM_ADM2 with 2021 source.


lydiatroup (approved):
This looks great! Thank you for the submission

PR #2387, opened 2022-09-08
PRI_ADM0 gbOpen

maxmalynowsky:

Create PRI_ADM0 with 2021 source.


lydiatroup (changes requested):
Please change the canonical boundary type name to reflect the name of the ADM (Puerto Rico). Thank you!

PR #2392, opened 2022-09-08
VIR_ADM1 gbOpen

maxmalynowsky:

Create VIR_ADM1 with 2021 source.


lydiatroup (changes requested):
Please re upload this file as an ADM0 (changing the name and level in the meta file, along with the level in the attribute table) in accordance with https://github.com/wmgeolab/geoBoundaries/discussions/2509

PR #2402, opened 2022-09-09
IND_ADM0 gbOpen

maxmalynowsky:

Update IND_ADM0 with 2011 source.
Created by dissolving state boundaries, so they are perfectly hierarchical.


lydiatroup (changes requested):
Please change the canonical boundary type name to "The Republic of India," as shown in the UN stats document discussed previously. This change should also be made in the attribute table of the file.

Thanks!

Issue #2050, opened 2022-01-30
gB Open IND ADM1


lydiatroup:
layer needs to be replaced

PR #2403, opened 2022-09-09
IND_ADM1 gbOpen

maxmalynowsky:

Update IND_ADM1 with 2011 source.
Old boundary has errors displaying Ladakh and Telangana.


lydiatroup (approved):
Looks great! Thank you for the contribution! closes #2050

PR #2404, opened 2022-09-09
IND_ADM2 gbOpen

maxmalynowsky:

Update IND_ADM2 with 2011 source.
Old boundary has dozens of sliver gaps and sliver polygons that cannot be automatically fixed.


lydiatroup (changes requested):
One small change needed, otherwise this file looks great. The level field in the attribute table is populated as "ADM1", but should be "ADM2".

PR #2407, opened 2022-09-09
FRA_ADM0 gbOpen

maxmalynowsky:

Update FRA_ADM0 with 2022 source.


lydiatroup (changes requested):
Please change the canonical boundary type name field in the meta data to accurately reflect the name of this country (The French Republic). This name comes from the un stats document discussed previously. Additionally, please change the "France" in the name field of your attribute table to the correct name.

Thank you!

Issue #2044, opened 2022-01-30
gB Open FRA ADM1


lydiatroup:
ADM1 missing polygon for Corsica Island, fix this and reupload

PR #2408, opened 2022-09-09
FRA_ADM1 gbOpen 

maxmalynowsky:

Update FRA_ADM1 with 2022 source.


lydiatroup (approved):
This file looks great! Thank you for the contribution. closes #2044

PR #2412, opened 2022-09-09
GLP_ADM0 gbOpen 

maxmalynowsky:

Create GLP_ADM0 with 2022 source.


lydiatroup (changes requested):
Please change the Canonical Boundary Type Name field in the meta file to reflect the name of this ADM. (Guadeloupe)

PR #2421, opened 2022-09-09
GUF_ADM4 gbOpen 

maxmalynowsky:

Create GUF_ADM4 with 2022 source.


lydiatroup (changes requested):
Please change the NOM field in the attribute table to "Name" and this file will look perfect

PR #2432, opened 2022-09-09
REU_ADM0 gbOpen 

maxmalynowsky:

Create REU_ADM0 with 2022 source.


dfornatora (changes requested):
Thank you for your contribution. For Canonical Boundary Type Name in the meta, please change from Territory to Réunion. After the changes, please reupload to this pull request.

PR #2433, opened 2022-09-09
REU_ADM1 gbOpen 

maxmalynowsky:

Create REU_ADM1 with 2022 source.


lydiatroup (changes requested):
Please reupload this file as an ADM0, meaning the canonical boundary name in the meta would need to be changed to the name of the territory and the boundary type changed to adm0.

PR #2434, opened 2022-09-09
REU_ADM2 gbOpen 

maxmalynowsky:

Create REU_ADM2 with 2022 source.


lydiatroup:
closing bc we will not include this layer with the new method of addressing territories (see the discussions for more information)

Issue #2291, opened 2022-06-21
[BOUNDARY ERRATA] The Gambia Central River admin 1 split is incorrect


cclarkson:
In The Gambia, the Central River admin 1 region has, in recent years, been split into two. It looks like this has been done horizontally, with Kuntaur being the new northern admin 1 region and Janjanbureh becoming the southern admin 1 region 
_e.g._ DOI:[10.1371/journal.pone.0139354](http://dx.doi.org/10.1371/journal.pone.0139354) and on The Gambia Bureau of Statistics [website](https://gambia.opendataforafrica.org/).

However, on wikipedia, I see some examples where the this division has instead been made vertically , creating an eastern Kuntaur and and western Janjanbureh, though I can see any "official" recognition of this. The data in geoboundaries follows this (incorrect?) division.

Additionally, the admin 1 shapeISO is identical for both Kuntaur and Janjanbureh, `GM-M`, relating the old Central River region. I would have thought that new shapeISOs would have been created when Central River was split, though I can't find any official new shapeISOs online.

DanRunfola:
Thank you for this submission!  Just as a note we're a bit slow over the summer (being based out of a university!) but we'll try to get this into the database ASAP.

kpupkiewicz:
Good spot, @cclarkson . You’re correct about Kuntaur & Janjanbureh - all official Gambian sources have that boundary division at the river border. I have no idea how that error on Wikipedia came about 🤔 Check out the metadata in the PR above for the updated source information. Hopefully tomorrow afternoon I can add a new GMB_ADM3 from this source as well.

PR #2439, opened 2022-09-10
GMB_ADM1 gbOpen

kpupkiewicz:

#2291 

## Anything Unusual?

shapeISO's have yet to be updated by ISO to accommodate the new administrative structure. Current submission uses the previous structure, and some LGA's (local government areas) will share shapeISO labels that applied to their previous divisions.


lydiatroup (changes requested):
A few necessary changes and this file will be good to go
1) change the license in the metadata to "Creative Commons Attribution 4.0 (CC BY 4.0)" and remove that from the license notes
2) Canonical boundary name appears to be region not local government areas

kpupkiewicz:
Good catch @lydiatroup !
I did get an email when I uploaded the new file, https://github.com/wmgeolab/geoBoundaries/actions/runs/3048104091 - I don't know enough about github to say whether or not this is important but I figured I'd make sure you saw it nonetheless.

lydiatroup (approved):
Thanks for the fix! closes #2291

lydiatroup:
@kpupkiewicz I get emails from each PR run so I did see that, but the file looks all good now so I went ahead and approved it!

PR #2440, opened 2022-09-11
gB Open JOR ADM0 #2009

ssheminway:
JOR ADM0

 #2009


lydiatroup (changes requested):
just two changed needed-
1) the range of dates is actually only needed for salb files, so I would just put 2016 for this one
2) there's two shape length columns in your attribute table - there should only be one so try and delete one of them! (one should be delete protected and one likely won't be)

lydiatroup:
@dfornatora

PR #2441, opened 2022-09-12
GMB_ADM2 gbOpen

kpupkiewicz:
## Anything Unusual?

Source included data at the ADM2 level, but it appeared to be improperly projected. This ADM2 level was generated by geoBoundaries by dissolving the source's ADM3 polygons by the 'District' field values. No other modifications were made.


lydiatroup (approved):
looks good!

Issue #2205, opened 2022-03-28
gB Open UGA ADM4 


lydiatroup:
in need of a file for Uganda adm4

shauntmathew:
Here is a link to an image to digitize the sub-counties (ADM4) but it is outdated.
https://commons.wikimedia.org/wiki/File:Uganda_sub-counties.png

PR #2446, opened 2022-09-12
UGA_ADM4 gbOpen

rmbiddle:
 #2205 2205


lydiatroup (changes requested):
A few fixes needed for this file:

1) license type should be "Creative Commons Attribution 3.0 Intergovernmental Organisations (CC BY 3.0 IGO)" and then the meta will be good
2) please make a new field in the attribute titled 'Name' and calculate it to = what is currently in the sname2019 field. just renaming the field won't work, you have to make a new field and calculate it
3) This file is in the wrong projection! please fix to be in EPSG:4326 - WGS 84 (WGS 1984 geographic coordinate system in arcgis pro)

let me know if you have any questions on those

lydiatroup:
@dfornatora

lydiatroup (approved):
looks great! closes #2205

PR #2451, opened 2022-09-13
ASM_ADM0 gbOpen

maxmalynowsky:

Current ADM0 is missing an island and does not align with proposed ADM1 from US Census Bureau
https://github.com/wmgeolab/geoBoundaries/pull/2449


lydiatroup (changes requested):
Please change the canonical boundary type name field in the meta file to reflect the name of the ADM (American Samoa). 
Thank you!

PR #2469, opened 2022-09-15
SLV_ADM0

taraamcl:
fixed gbBot pr


lydiatroup (changes requested):
you're missing a couple files in the zipfile and that's why it's failing; there should also be a .shp, .shx, .sbn file (and maybe more..). I'd re export your file to get those required files or just rezip ALL the files from your export!

Issue #2292, opened 2022-07-05
[BOUNDARY ERRATA] [FRA] Missing overseas territories for FRA ADM0 and ADM1


ibhalin:
French overseas territories are missing for ADM0 and ADM1 levels but are present from ADM2 onward. 

**ADM0**
![image](https://user-images.githubusercontent.com/106236433/177273258-f838bca4-2dbb-4070-af0a-c318fe97635d.png)

**ADM1**
![image](https://user-images.githubusercontent.com/52789653/177274566-a88b43c3-851b-4667-994d-6ea507897be2.png)

**ADM2**
![image](https://user-images.githubusercontent.com/52789653/177274658-07d29d58-ccbc-4827-980a-558a4a834fed.png)

DanRunfola:
Thank you for reporting this, very helpful - we'll look into it as soon as we can (note that we're based out of a US academic institution, so it may be late August before we're fully back, so apologies if it takes a bit!)

DanRunfola:
We went around on this a *bunch*, and where we've temporarily landed is that we're going to only include territories in the parent country at the ADM levels for which the parent country defines htem.  I.e., France does not actually include these territories in it's ADM0 as per the sources we have right now, but it does at ADM3.

...I think.  To be honest this has gotten pretty complicated, so I would welcome discusison over at https://github.com/wmgeolab/geoBoundaries/discussions/2509 if this solution is suboptimal for any reason.  We have a few colliding territory issues here, so I want to keep the conversation over there (closing this issue accordingly, but with the recognition we might re-open if warranted).

PR #2492, opened 2022-09-19
PYF_ADM0 gbOpen

kpupkiewicz:

Found a high quality source for French Polynesnia while working on #2292 . 

## Anything Unusual?

French Polynesia is a COM, one of five French overseas collectivities. It is also recognized uniquely as a POM (FR: Pays d'outre-mer, EN: Overseas Country). French Polynesnia should thus be included at all levels of FRA data.


DanRunfola:
Just noting we already had a PR submitted which was equivalent to this, sorry :)

Issue #2444, opened 2022-09-12
[BOUNDARY ERRATA] outdated NLD ADM1


BrentWHR:
The boundry of the province Utrecht has changed in 2019 due to municipal redistribution. NLD ADM2 is probably outdated as well.

The screenshot show the approximate new border given by GADM in 2022
![afbeelding](https://user-images.githubusercontent.com/72964508/189681150-20ba005f-746d-45f4-95a3-8e2beb80ffec.png)

The new provincial borders may be derived from:
https://nationaalgeoregister.nl/geonetwork/srv/dut/catalog.search#/metadata/e73b01f6-28c7-4bb7-a782-e877e8113e2c?tab=general

The provincial boundry was changed with a reordering of municipalities: https://en.wikipedia.org/wiki/Vijfheerenlanden

BrentWHR:
On second thought; this may be a better source file containing ADM0, ADM1, and ADM2:
https://nationaalgeoregister.nl/geonetwork/srv/dut/catalog.search#/metadata/208bc283-7c66-4ce7-8ad3-1cf3e8933fb5?tab=relations

DanRunfola:
I just wanted to pop in and say thanks for this!  We're ramping up for the academic year here, so hopefully can engage with this one sooner than later.

PR #2497, opened 2022-09-20
NLD_ADM0 Issue #2444

dfornatora:
 /issues/2444


lydiatroup (commented):
kingdom of the netherlands!

Issue #2493, opened 2022-09-20
gB Open ARG ADM0


lydiatroup:
Need a new file for ADM0 of argentina

PR #2498, opened 2022-09-21
ARG_ADM0 Issue #2493

dfornatora:
 /issues/2493


lydiatroup (changes requested):
This file still has adm1 information in the metadata. otherwise it looks good!

lydiatroup (approved):
looks great closes #2493

Issue #2010, opened 2022-01-27
gB Open JOR ADM1


lydiatroup:
need a new file for JOR

SidonieH:
https://www.just.edu.jo/FacultiesandDepartments/FacultyofEngineering/Departments/CivilEngineering/Pages/gis_maps.aspx#

PR #2511, opened 2022-09-23
JOR_ADM1 Issue #2010

dfornatora:
 /issues/2010


lydiatroup (approved):
looks great! thanks for the submission!

PR #2513, opened 2022-09-26
JPN_ADM1 #1795

lydiatroup:


lydiatroup:
@shauntmathew You'll want to project this file to the correct coordinate system (WGS 1984!) and license type to "Creative Commons Attribution 4.0 International (CC BY 4.0)" and then try uploading it to this PR with JUST the adm1, and you may be able to avoid having to do lfs

Issue #2053, opened 2022-01-30
gB Open IND ADM4


lydiatroup:
layer needs to be replaced

PR #2529, opened 2022-10-01
IND_ADM4 #2053

hadleyday:
IND_ADM4

 https://github.com/wmgeolab/geoBoundaries/issues/2053


lydiatroup (changes requested):
two fixes needed:
1) boundary type in meta data should be 'ADM4' not just '4'
2) your shapefile is missing the .prj file, make sure you get them all in the zipfile

leeberryman:
@lydiatroup also cannonical boundary for ADM4 is CD Block not sub district.  ADM3 is Tehsil / Taluka = subdistrict.

Issue #2018, opened 2022-01-28
gb Open AGO ADM2


lydiatroup:
Country names are in all caps. Correct to title casing and correct missing character symbols "?"

casing can be fixed using the field calculator [in arc](https://support.esri.com/en/technical-article/000011159)

PR #2531, opened 2022-10-02
#2018 AGO_ADM2

VictorGedeck:
Fixed boundary data for AGO_ADM2 (#2018)

 https://github.com/wmgeolab/geoBoundaries/issues/2018

## Anything Unusual?
 It uses the most recent licensable open-source data however it is not fully reflective of the current ADM2 boundaries in Angola as there have been changes since the boundary data was put on the internet.


lydiatroup (changes requested):
Two changes and this file will be good to go:
1) in the meta it should be "ADM2" not "ADM 2"
2) please delete the ISO_Code column - it is not needed for ADM2s and above

lydiatroup (changes requested):
File itself looks good, but is named "AGOADM_2" 
please fix that to be AGO_ADM2 and upload it to a NEW pr (have to close this pr and create a new one when a file is named wrong or uploaded to the wrong location)

PR #2533, opened 2022-10-03
#2018 AGO_ADM2

VictorGedeck:
 https://github.com/wmgeolab/geoBoundaries/issues/2018

## Anything Unusual?
 It uses the most recent licensable open-source data however it is not fully reflective of the current ADM2 boundaries in Angola as there have been changes since the boundary data was put on the internet.


lydiatroup (approved):
this looks great!! amazing job digitizing - closes #2018

Issue #2360, opened 2022-09-07
gB Open PAN ADM0


lydiatroup:
We need a new file for PAN ADM0!

leeberryman:
@lydiatroup does this meet the need for PAN_ADM0?

DanRunfola:
Closed by https://github.com/wmgeolab/geoBoundaries/pull/2542

PR #2535, opened 2022-10-04
PAN_ADM0#2360

JialuT:
 #2360


DanRunfola:
@all-contributors add @JialuT  to data

Issue #1857, opened 2021-09-23
gB OPEN SVN ADM1


lydiatroup:
need a new file for SVN ADM1 (no humdata, share-alike, or odbl)

hadleyday:
https://international.ipums.org/international/gis_harmonized_1st.shtml

hadleyday:
https://ec.europa.eu/eurostat/web/gisco/geodata/reference-data/administrative-units-statistical-units/nuts#nuts21

PR #2538, opened 2022-10-05
SVN_ADM1 #1857

hadleyday:
SVN_ADM1

 https://github.com/wmgeolab/geoBoundaries/issues/1857


lydiatroup (changes requested):
this file has the india meta data file in it! please add the correct meta and delete this meta

lydiatroup (approved):
looks good! for future reference, try to get the url in the license image. closes #1857

Issue #2219, opened 2022-03-31
gB Open MMR ADM1


lydiatroup:
Either can correct the meta if the current file has a compatable license, otherwise find a new file and replace it!

gbOpen for Myanmar ADM1 appears to have inccorect source, link, and year metadata. Both list OpenStreetmap in the boundarySource-1 and licenseSource fields. However, boundarySourceURL links to http//geonode.themimu.info/layers/?limit=5&offset=10, which is simply a list of several Myanmar related admin datasets, most of which are much more recent than 2014 as listed in the boundaryYear field.

ssheminway:
Couldn't find a map to georeference without a share alike license

PR #2545, opened 2022-10-09
ISSUE #2219 MMR ADM1

MichaelR-geoLab:
dissolved from ADM 2 source and fixed holes


lydiatroup (approved):
amazing work as always! closes #2219

PR #2547, opened 2022-10-12
IRL_ADM1

hadleyday:
IRL_ADM1

 Current IRL_ADM1 file uses data from OSM


lydiatroup (changes requested):
Your shapefile has some ring self intersects in it that are causing the bot to fail. You may be able to fix that in map shaper or by editing some of the vertices in the file. Also there's a few extra columns in the attribute table so you'll want to delete those. 
And when you make these changes upload to a new PR and be sure to put it in the source data / gb open folder!

lydiatroup (changes requested):
You have two different IRL_ADM1s in this PR - one uploaded to sourceData and one uploaded to sourceData/gbOpen (this is the correct location). Please open a new PR with just the corrected file and I think we should be good!

Issue #2033, opened 2022-01-30
gB Open CRI ADM3


lydiatroup:
ADM3 is missing polygons, find a new file or fix this error

PR #2550, opened 2022-10-17
#2033 CRI_ADM3

jingjli1:
fixed missing polygons


lydiatroup (changes requested):
Just updated the issue for this, but we can't use anything from OSM because of the license (share-alike / oDbL). You'll need to find a new file for this issue!

lydiatroup:
closing bc wrong license type

Issue #1860, opened 2021-09-23
gB OPEN SWE ADM2


lydiatroup:
need a new file for SWE ADM2 (no humdata, share-alike, or odbl)

PR #2552, opened 2022-10-18
gbOpen SWE_ADM2 #1860

amanda-reed:
 #1860 

## Anything Unusual?

I had to use 2 sources to digitize. The second license PNG is included in the zip as well as the additional meta data in the meta.txt file.


lydiatroup (approved):
looks great! closes #1860

PR #2555, opened 2022-10-21
gB Open CHN ADM1

JialuT:
#2028

## Anything Unusual?

The disputed areas are in a small part of southwestern Xinjiang Uyghur Autonomous Region, western and southern Tibet Autonomous Region, and Taiwan Province


amanda-reed (changes requested):
Just two quick fixes I noticed in the metadata: if its digitized, source 1 should be geoBoundaries, and then source 2 would be Wikimedia Commons; it looks like the link to source data is for Panama so just make sure you update that as well with the right link. Other than that, everything looks great!

amanda-reed (approved):
Looks great! closes #2028

Issue #1995, opened 2022-01-13
[BOUNDARY ERRATA] LUX ADM2 & ADM3


leeberryman:
As I'm working on the canonicalBoundary labels I cam accross what might be an anomaly in the geoboundaries that should be looked into. In Luxembourg are the ADM2 & ADM3 files both valid? one has 105 features and the other has 102 features...  ADM3 looks like it comes from the LUX government and the ADM2 is from OSM.

PR #2556, opened 2022-10-23
LUX_ADM2-3 gbOpen

rmbiddle:
 #1995


lydiatroup (changes requested):
a few of these files are uploaded to the wrong place and I think that's causing some issues with the bot. Make a new pr with the two files in the right place (sourceData/gbOpen) and then we can see what any problems are!

It looks like the file uploaded just as LUX_ADM3 is good but the one uploaded to sourceData/gbOpen is not, so make sure you're using the zipfile will all the files in it for your new pr!

Issue #2327, opened 2022-08-10
[BOUNDARY ERRATA] Guatemala ADM2


DanRunfola:
**Describe the Error**
GTM ADM2 has some holes - not sure if it's in our build or in the source, but we need to check.

**Screenshots**
![image](https://user-images.githubusercontent.com/7882645/183909589-58ca00a0-52c1-4337-b218-4a43f89ab727.png)

lydiatroup:
the top right hole is also present in the source (https://data.humdata.org/dataset/cod-ab-gtm) and is a lake. some research into this indicates the lake itself may not be split into municipalities

MichaelR-geoLab:
Should I make the holes into separate provinces, split them among existing provinces, or leave it be? Not sure about the procedure here.

lydiatroup:
@MichaelR-geoLab It'll depend on which hole it is. Will require some research to see how they are actually split up ie whether it's split or all in one or its own thing. If you don't see anything promising after some research, I would fill in the holes as their own features with them listed as the lake name.

PR #2560, opened 2022-10-28
Issue #2327 GTM ADM2

MichaelR-geoLab:
Found a source which said the lakes are their own ADM2 features and added them.


lydiatroup (changes requested):
This file looks good except we no longer can use anything on humdata/hdx as a source (or OCHA). Is that where the shapefile came from or where you got the information about the lakes?

lydiatroup (approved):
we had talked about this in class and determined it was okay - closes #2327

PR #2561, opened 2022-10-28
IRL_ADM1

hadleyday:
IRL_ADM1

 Current ADM1 issue is incorrect


lydiatroup (changes requested):
One more fix and this will be good! Delete the extra columns in the attribute table (shape id and shape len 1)

amanda-reed:
@hadleyday Just to double check, this PR is not linked to a specific issue # but fixes a known issue for IRL ADM1?

lydiatroup (approved):
looks good!

PR #2564, opened 2022-10-31
LUX_ADM3 gbOpen

rmbiddle:
 #1995


amanda-reed (changes requested):
Similar to LUX_ADM2, source 1 should be geoBoundaries, and source 2 should be Wikimedia Commons, but since OpenStreetMap was credited, I'm not sure if we can use this source. Also, the license source should have the link where the license is indicated, so the same link as source in this case. I'll check with Lydia about OSM and get back to you by class tomorrow.

PR #2565, opened 2022-10-31
LUX_ADM2 gbOpen

rmbiddle:
 #1995


amanda-reed (changes requested):
Hi Russ, similar to MLT_ADM2, there's just a few quick changes needed in the metadata: 
- since it's digitized, source 1 should be geoBoundaries and source 2 should be Wikimedia Commons
- according to your source, the license should be CC0 1.0 Universal (CC0 1.0) Public Domain Dedication, which is different from just Public Domain
Otherwise, looks good!

Issue #1849, opened 2021-09-23
gB OPEN IRQ ADM1


lydiatroup:
need a new file for IRQ adm1 (no humdata, share-alike, or odbl)

PR #2568, opened 2022-11-03
IRQ_ADM1 gbOpen

rmbiddle:
 #1849


amanda-reed (changes requested):
2 quick changes: the bot is sorta picky so the license should be 'CC0 1.0 Universal (CC0 1.0) Public Domain Dedication' and then License source should be the link where the license is specified, so in this case it's the same as the data source link. Everything else looks good!

Issue #2034, opened 2022-01-30
gB Open CRI ADM2


lydiatroup:
current file is incorrect, need a new file

PR #2570, opened 2022-11-06
CRI_ADM2 #2034

VictorGedeck:
CRI_ADM2
https://commons.wikimedia.org/wiki/File:Mapa_de_cantones_de_Costa_Rica.svg

 https://github.com/wmgeolab/geoBoundaries/issues/2034


amanda-reed (changes requested):
Hi Victor, just one quick fix in your metadata: the license should be 'CC0 1.0 Universal (CC0 1.0) Public Domain Dedication' instead of just public domain as specified in the source link. Other than that, everything looks good!

Issue #1839, opened 2021-09-22
gB Open HUN ADM2


lydiatroup:
need a new file for HUN adm3 (no humdata, share-alike, or odbl)

PR #2571, opened 2022-11-08
HUN_ADM2 #1839

taraamcl:
HUN_ADM2 #1839

Budapest is called the capital city and not a county so I'm not totally sure how to address that. It's surrounded by the county of Pest but I'm not sure if Pest includes Budapest? I think they are administered separately so I added Budapest as its own entry. Just a heads up!


lydiatroup (commented):
your budapest region actually is not showing up as a separate feature on the map - i'm thinking this is just a gis issue & we can mess with it in class tomorrow!

lydiatroup (changes requested):
Since this is an adm2, we don't need the ISO_Code column in the attribute table so please delete that! Also the ID field is all zeros so if you can, delete that aswell. also for level in the attribute table is 'ADM2' not 'ADM_2'

lydiatroup:
@taraamcl

Issue #2586, opened 2022-11-11
gB OPEN BGR_ADM2


lydiatroup:
Need a new file for BGR ADM2, current source is OSM

PR #2594, opened 2022-11-13
BGR_ADM2 Issue #2586

dfornatora:
 /issues/2586


amanda-reed (approved):
Thanks for your hard work on this one! closes #2586

Issue #2587, opened 2022-11-11
gB OPEN BHR ADM0


lydiatroup:
Need a new file for BHR ADM0, current source is OSM

christillotson:
Will be corrected as soon as #4229 is merged.

PR #2601, opened 2022-11-14
BHR_ADM0 #2587

taraamcl:
BHR_ADM0 #2587


lydiatroup (changes requested):
we can't use this file because of the license type - it is creative commons share alike, and we can't use anything with 'share alike' in the name. see this doc for more info: https://docs.google.com/document/d/1lGvDZlDr0KbR8peDY9MkLjjT2BUJFBubB9p4JbbybdM/edit

Issue #2588, opened 2022-11-11
gB OPEN BHR ADM1


lydiatroup:
Need a new file for BHR ADM0, current source is OSM

christillotson:
Will be corrected once #4230 is merged

PR #2602, opened 2022-11-14
BHR_ADM1 #2588

taraamcl:
BHR_ADM1 #2588


lydiatroup (changes requested):
same thing as the ADM0 - we can't use anything 'share-alike'

Issue #2582, opened 2022-11-11
gB OPEN BGD_ADM0


lydiatroup:
Need a new file for BDG ADM0, current source is OSM

dfornatora:
Think this ISO Code might actually be BGD?

PR #2603, opened 2022-11-14
BDG_ADM0 #2582

taraamcl:
BDG_ADM0 #2582


lydiatroup:
There was a small geometry error in this one that I fixed and reuploaded, but FYI these ISO codes are also BGD not BDG (i typed them in wrong in the issue - my bad!)

Issue #1994, opened 2022-01-13
[BOUNDARY ERRATA] JOR ADM2


leeberryman:
looks like there is a character encoding issue in the JOR ADM2 source file.  Its also an OSM file so its probably already marked for replacement, but name encoding is what flagged it for me.

PR #2621, opened 2022-11-16
issue #1994 JOR_ADM2

VictorGedeck:
JOR_ADM2
 https://commons.wikimedia.org/wiki/File:Jordan_nahias.png

 https://github.com/wmgeolab/geoBoundaries/issues/1994

## Anything Unusual?
 Written in metadata: Source is slightly outdated as there have been changes in districting since 2006 however this is the most recent source that has a license we can utilize.


lydiatroup (approved):
looks great! close #1994

PR #2634, opened 2022-12-05
Add files via upload

jshobart:


lydiatroup (changes requested):
Two more changes that I missed before! For digitized files source 1 is geoBoundaries and source 2 is where the image came from, so just switch your source 1 and source 2 in the meta. Also, the 'ISO' column in the attribute table should actually be named 'ISO_Code'. To fix that you'll need to make a new column and populate it with the iso codes and then delete the one just called 'ISO' (you can't rename a column).

PR #2639, opened 2022-12-06
ECU_ADM0

shauntmathew:
ECU_ADM0

 Old ECU_ADM0 is from Open Street Map, this is digitized from the same map as the ECU_ADM1


lydiatroup (changes requested):
your meta file still has 'Boundary Type: ADM1', so switch that to ADM0. Also in the other notes, you would say it was 'dissolved' from ADM1, not 'merged'.

PR #2640, opened 2022-12-07
Add files via upload

jshobart:


lydiatroup (changes requested):
thanks for making that change Jess! Still one more- the ISO column in the attribute table should be named 'ISO_Code'. to do this you'll need to make a new column and delete the current one! (you can't rename a column in arcgis pro)

Issue #2042, opened 2022-01-30
gB Open FJI ADM3


lydiatroup:
need a new file for this adm, current file as errors and has an incompatible license

PR #2645, opened 2022-12-10
FJI_ADM2 #2042

Sarangaban:


amanda-reed (changes requested):
Just one quick fix in the metadata: the license should be 'CC0 1.0 Universal (CC0 1.0) Public Domain Dedication' which is different from just Public Domain.

Issue #2280, opened 2022-05-07
[BOUNDARY ERRATA - Pakistan: Balochistan, Sindh out of date by over twenty years; Azad Kashmir's districts not displayed]


Caneman786:
Balochistan's districts are out-of-date by over twenty years, dating to the 1998 census.
Here is a map dating to 2021 (missing newly-formed Chaman) provided by the provincial authority:
https://cmdu.gob.pk/wp-content/uploads/2019/11/Districts-scaled-1-1024x724.jpg

Sindh's districts are out-of-date by over twenty years, dating to the 1998 census.
Here is a map dating to 2017 (missing newly formed Kemari) provided by the Pakistan Statistics Bureau:
https://i.imgur.com/w1ZNAb6.jpg

Azad Kashmir's districts are not even displayed, it has ten.
Here is an up-to-date map provided by the Pakistan Bureau of Statistics:
https://i.imgur.com/msHAZQC.jpg

**Additional context**
I think it would be helpful to prioritize this issue, which affects areas holding over 60 million people. Many people would be very happy to see and would be able to take advantage of a proper dataset with Pakistan's division and district boundaries.

DanRunfola:
Hi Caneman,

Thank you for reporting this - it is very much appreciated.  Please be aware that we (geoBoundaries) is based at an academic institution, and thus our cadence of updates during summer months is generally very slow, so it may be a few months before we can engage in earnest.  But don't let that in any way suggest that we don't highly value this contribution, and I look forward to when we can correct the errors.

Best,
Dan

MichaelBrandonFalk:
This may be an easy fix [https://data.humdata.org/dataset/cod-ab-pak](https://data.humdata.org/dataset/cod-ab-pak)

PR #2648, opened 2022-12-12
PAK ADM4

kpupkiewicz:

#2280


amanda-reed (changes requested):
Hi Kristina, it looks like the license at the source link is actually CC Attribution Share-Alike International 4.0, and we can't use any share-alike licenses. Sorry for the finicky licensing rules, but you would have to find a new source for this issue.

DanRunfola:
@kpupkiewicz Can you please re-upload this one?  It got closed due to the need to refresh the geoBoundaries repository during the 5.0 release!

PR #2684, opened 2023-02-06
Example PR GRC_ADM3

amanda-reed:


amanda-reed:
It took a while to clone the repository so I think it cloned the whole thing rather than just the pointers. I was looking around based on the bot error and found this but I'm not sure if it's relevant: https://github.com/SamKirkland/FTP-Deploy-Action/issues/77

amanda-reed:
Made a mistake and inserted GRC_ADM3 into itself rather than replacing it, new PR opened

Issue #2116, opened 2022-02-11
gB Open SLE ADM 0-4


lydiatroup:
getf has requested we upload new files for all of these layers

only upload if the file is more recent than the dates below
<img width="407" alt="Screen Shot 2022-02-11 at 2 52 34 PM" src="https://user-images.githubusercontent.com/60264585/153660126-9341c180-0651-4374-b3ad-7fa1ad820edd.png">

kpupkiewicz:
The last census in Sierra Leone was completed in 2015, so most shapefiles are from around that time period. The country did some redistricting in 2017 so any new resources need to be from after 2016 at the earliest. 
Considering this, digitizing seems to be the best option. Do we have any new standards for digitizing @lydiatroup ? Is wikipedia/wikimedia OK to use? This [wikipedia page](https://en.wikipedia.org/wiki/Subdivisions_of_Sierra_Leone) uses the post-2017 organization, and it seems to be of appropriate detail.

lydiatroup:
@kpupkiewicz for digitizing we're looking both for accuracy (the image picked matches the right number of areas and shapes) and the license type! For license type you can use anything that is open source EXCEPT licenses that include "share-alike" in the title or OBDL licenses (osm). If you click on the images on that wiki page it'll pop up with the license type. it looks like only the last one (the chiefdoms) are of an appropriate license type

kpupkiewicz:
@lydiatroup What do we think of the viability of this data? https://nec.gov.sl/site/boundary It is the last link on the page, called 'National Electoral Commission (NEC), Ward Boundary Delimitation Report, 2017'. As far as I can tell it contains data for ADM 2 (districts), ADM 3 (chiefdoms), and ADM 4 (wards). It would also be possible to create ADM1 data from what is in the file. However, the data itself is of pretty poor quality. Also, I am not sure what kind of license would apply in this case.

ejott:
ADM1: #4232 
ADM2: #4231 

ADM3: could not find any up-to-date sources
ADM4: Wards information can be found [here ](https://ec.gov.sl/wards/)to start, and this [file ](https://ecslmedia.ec.gov.sl/pdfviewer/bd-constituency/?auto_viewer=true#page=2&zoom=auto&pagemode=bookmarks) has maps for reference, but the maps are very distorted and there is no license information to be found. 

Given the difficulty of finding data for ADM3 and ADM4, it might not be possible to complete this issue completely.

PR #2767, opened 2023-02-24
"SLE_ADM3 Issue #2116"

MichaelR-geoLab:


amanda-reed (changes requested):
Hi Michael, I believe there's some really small holes in your boundaries which you can fill with "delete holes" in qgis or I'm sure you can do it in arc as well. Also, make sure your license picture includes the source link and that you switch the source order in meta so geoBoundaries is source 1. If filling the holes doesn't fix the issue, I'll look more into it!

Sarangaban:
New issue created.

Issue #2755, opened 2023-02-12
gB Open ISL ADM0


lydiatroup:
need a new file for ISL ADM0

PR #2774, opened 2023-03-02
ISL_ADM0 #2755

ejott:


lydiatroup (changes requested):
This was a great find! only one change to make- 
the "Canonical Boundary Type Name:" in the meta.txt file should be "Iceland" not "island"
please make that change and reupload

lydiatroup:
@ejott I have no idea why this is failing- it looks good to me. Could you try uploading as a new PR when you get a chance?

Issue #2772, opened 2023-02-26
gB Open LVA ADM0


lydiatroup:
need a new file for LVA ADM0

PR #2775, opened 2023-03-02
LVA_ADM0 #2772

ejott:


lydiatroup (changes requested):
Another good find! a few changes needed and file will be good to go:

- canonical boundary type name should be "the Republic of Latvia" (this comes from this [site](https://www.iso.org/obp/ui/#iso:code:3166:LV)) 
- this means you'll also have to change the name under "name in the attribute table

I'll also create an issue for LVA adm1 and assign you to it, if you could complete that issue with the same data source that would be great!

lydiatroup (approved):
looks good! closes #2772

Issue #1659, opened 2021-07-30
gB OPEN LVA_ADM2


sgmurphy216:
Need a new source for LVA_2; current is very incomplete

PR #2776, opened 2023-03-02
LVA_ADM2 #1659

ejott:

### DIFF
--- a/sourceData/gbOpen/LVA_ADM2.zip
+++ b/sourceData/gbOpen/LVA_ADM2.zip
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:fe7d41dfebbe72ece9f468f1ac8f87e52c0e6fa06485c2c9b88bc8125417279b
-size 3264300
+oid sha256:eab47c4df4f33e1ca5e7cd17a2b026550219f55246f16007456c99d4cc82c75f
+size 3264295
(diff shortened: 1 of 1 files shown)


lydiatroup:
will also address 1694

lydiatroup (changes requested):
only one fix needed: if the file was NOT digitized please remove geoboundaries from source 1. otherwise, disregard this

ejott:
> only one fix needed: if the file was NOT digitized please remove geoboundaries from source 1. otherwise, disregard this

Gotcha, I did have to change it (and LVA_ADM0) from .gml to .shp which is why I included that b/c that was a bit of a process for me but thinking back that's not really digitization, is it? will fix and remove

lydiatroup:
> > only one fix needed: if the file was NOT digitized please remove geoboundaries from source 1. otherwise, disregard this
> 
> Gotcha, I did have to change it (and LVA_ADM0) from .gml to .shp which is why I included that b/c that was a bit of a process for me but thinking back that's not really digitization, is it? will fix and remove

A little more info on this for you. The reason we put geoboundaries as source one for digitization is because we are then CREATING the file itself. In cases like yours or when dissolving a boundary, making a few minor edits, you are editing the existing as opposed to creating it from scratch, so geoBoundaries isn't a source. In essence, this 'source' is asking for where geoBoundaries found the file!

Issue #2676, opened 2023-02-03
gb Open CHL ADM0


lydiatroup:
need a new file for CHL ADM0

justinelliotmeyers:
chl chile? I can help if so 'eh

leeberryman:
@justinelliotmeyers yes.  Do you have a link to data with clear license?

justinelliotmeyers:
https://www.ine.gob.cl/herramientas/portal-de-mapas/geodatos-abiertos
https://www.ine.gob.cl/terminos-de-uso-y-licencia-de-datos-abiertos


Adios!

PR #2780, opened 2023-03-03
CHL ADM0 Update #2676

jamesnelmore:


amanda-reed (changes requested):
Hi James -- these boundaries look great but unfortunately we can't use anything from HumData. This is tricky because the source link is Amerigeo but the download link goes to humdata, so you'll need to find a new source for this issue. Let me know if you have any questions about sources - so sorry!

Issue #2249, opened 2022-04-25
Russia ADM2 - column name show as question marks `?????`


Rub21:
**Describe the Error**
Russia  admin-2 column name show as incorrect! The name column show in question marks `?????`

Source:  https://github.com/wmgeolab/geoBoundaries/blob/main/sourceData/gbOpen/RUS_ADM3.zip?raw=true

**Screenshots**
![image](https://user-images.githubusercontent.com/1152236/165105691-8644f5f7-4572-40cb-8736-447f446e5bef.png)

**Additional context**
Russia  admin-0 and admin1 show  in a correct way.

![image](https://user-images.githubusercontent.com/1152236/165106337-682f90d2-7e22-40f0-a01b-e39d971a2cb3.png)
![image](https://user-images.githubusercontent.com/1152236/165106384-e3aba53e-fee6-416e-8be9-23049b09f615.png)

cc. @yunica @geohacker

lydiatroup:
Thank you for finding this bug for us! We will try to fix this issue asap!

dfornatora:
After looking into this, there is no available data with the appropriate license we can use. [https://maps.princeton.edu/catalog/stanford-td486fb7677 ](https://maps.princeton.edu/catalog/stanford-td486fb7677) has information at this administrative level with the names, but it is unable to be used in our dataset.

DanRunfola:
@dfornatora Do we have the original data anywhere?  I.e., it looks like our ADM2 had the names, and they got scrapped due to some sort of encoding issue.

lydiatroup:
@dfornatora @DanRunfola https://drive.google.com/file/d/1RhrFw-_6QvLmneAKABuOxy8vI_GFMuQ-/view?usp=sharing
the link above has the original dataset. It is an incompatable license type but I could see the value in uploading it with an outdated license type that still has the names

PR #2781, opened 2023-03-03
RUS_ADM2 #2249

amanda-reed:
 #2249

Submitted by @dfornatora 
The license is one that we don't usually accept anymore, but since there is no data available to update, this is the original dataset so there is no longer an encoding issue with the Name attribute.


lydiatroup (approved):
approving this, thanks for reuploading this amanda! closes #2249

Issue #2666, opened 2023-02-03
gB OPEN BIH ADM0


lydiatroup:
need a new file for BIH ADM0

PR #2783, opened 2023-03-10
BIH_ADM0 #2666

amanda-reed:
 #2666 

Submitted on behalf of @hadleyday


lydiatroup (changes requested):
canonical boundary type name in the meta should be Bosnia and Herzegovina instead of 'Country'. Otherwise, looks good!

jingjli1:
close, new file was submitted.

Issue #2670, opened 2023-02-03
gB Open BRA ADM1


lydiatroup:
Need a new file for BRA ADM1

justinelliotmeyers:
brazil? lmk 'eh

leeberryman:
@justinelliotmeyers yes.  Do you have a link to data with clear license?

PR #2793, opened 2023-03-24
BRA_ADM1 #2670

jamesnelmore:
- #2670 BRA_ADM1
- fixed meta

### DIFF
--- a/sourceData/gbOpen/BRA_ADM1.zip
+++ b/sourceData/gbOpen/BRA_ADM1.zip
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:74f3c22b0663a81a864140463a39c06141addc434f922a5b5f7165c0f58c2d2b
-size 147290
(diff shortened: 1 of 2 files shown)


lydiatroup (changes requested):
A few changes needed for this file:
1) the name of the shapefile itself should be titled BRA_ADM1
2) source 2 in the meta data should be "Wikimedia Commons" because that is where the digitized image came from

also a note for the future, the automated digitizing is definitely cool, but you will need to smooth out the edges for file submissions from it (ie making them not as jagged / pixelated on the feature's edges)

lydiatroup (changes requested):
A few changes are needed:
- Source 1: should be geoBoundaries and source2: should be Wikimedia Commons
- Your attribute table has a few extra columns. Please remove the InPoly_FID, Shape_Le_1, and Shape_Area fields
- In the attribute table your ISO_Code and Level columns are flipped. ISO_Code should be the BR- values, while Level should be "ADM1"

Also great work on smoothing out the lines! this looks really good.

jamesnelmore:
I've made those changes. Thank you!

lydiatroup (approved):
looks great!! thanks! closes #2670

Issue #2671, opened 2023-02-03
gB Open BRN ADM0


lydiatroup:
need a new file for BRN ADM0

PR #2801, opened 2023-04-01
BRN_ADM0 #2671

VictorGedeck:


lydiatroup (changes requested):
a couple changes needed and then this file will be good!
1) source 1 in the meta should be geoBoundaries (bc you digitized it) & source 2 should be Wikimedia Commons. if you didn't digitize then ignore this!
2) I don't think this file is projected correctly, so please project it into WGS 1984 and reexport

Issue #2025, opened 2022-01-28
gB Open CAN ADM0-3


lydiatroup:
need a new file for these adms

maxmalynowsky:
Addressed in following:

https://github.com/wmgeolab/geoBoundaries/pull/2395
https://github.com/wmgeolab/geoBoundaries/pull/2396
https://github.com/wmgeolab/geoBoundaries/pull/2397
https://github.com/wmgeolab/geoBoundaries/pull/2398

lydiatroup:
@hadleyday can you knock out the adm3 for this issue so we can close it?

PR #2811, opened 2023-04-05
CAN_ADM0 #2025

amanda-reed:
 #2025 

uploaded on behalf of @hadleyday


lydiatroup (changes requested):
unsure who you uploaded this for, but please pass on the message below:

This canada file includes water boundaries (which we do not include in our dataset). IE we don't want whats in brown, we want what is in tan.
Screen Shot 2023-04-06 at 9 16 54 AM

One option to fix this is to take the ADM1 file from statcan and then dissolve it to make adm0

Sarangaban:
New file has been submitted without any issues.

PR #2812, opened 2023-04-05
CAN_ADM1 #2025

amanda-reed:
 #2025 

submitted on behalf of @hadleyday


lydiatroup (changes requested):
Only half of the features have an iso code filled in in the attribute table, please fill in the rest and reupload!

lydiatroup (approved):
Looks good!

PR #2819, opened 2023-04-06
CAN_ADM4 #2797

amanda-reed:


lydiatroup:
closing this because it looks like the Census Subdivision level does not have administrative power (and therefore isnt an admin boundary to include in the database!)

Issue #2678, opened 2023-02-03
gb Open CHN ADM0


lydiatroup:
need a new file for CHN ADM0

justinelliotmeyers:
China?

PR #2821, opened 2023-04-07
CHN ADM0 Issue #2678

MichaelR-geoLab:
- SLE_ADM3 update
- CHN_ADM0 update (Issue #2678)


MichaelR-geoLab:
Something is wrong with this PR, for some reason it is including SLE_ADM3 when I only included CHN_ADM0. Will try and fix.

amanda-reed (changes requested):
Is there a chance you had staged a commit for SLE_ADM3 #2767 and never pushed the commit? Instead of git add -A . you can manually add the files you want to commit with git add [staged file name]. If that doesn't work, I'll keep looking into it!

PR #2822, opened 2023-04-07
CHL_ADM0 #2676

jamesnelmore:
- #2670 BRA_ADM1
- fixed meta
- fixed title
- fixed meta
- Smoothed boundary and fixed metadata
- Updated CHL_ADM0 boundary

### DIFF
--- a/sourceData/gbOpen/BRA_ADM1.zip
+++ b/sourceData/gbOpen/BRA_ADM1.zip
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:74f3c22b0663a81a864140463a39c06141addc434f922a5b5f7165c0f58c2d2b
-size 147290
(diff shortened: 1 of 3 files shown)


amanda-reed (changes requested):
Hi James, just a few quick changes in the metadata for BRA_ADM1: if its digitized, source 1 should be geoBoundaries and source 2 should be the image source (Wikimedia Commons). In the attribute table, the only columns that we need are Name, Level, and ISO_Code, so you can get rid of the others before you export. Also make sure your projection is EPSG:4326 - WGS 84 (which is the reason why it's failing the bot). 

For CHL_ADM0, you seem to be missing the Canonical Boundary Type Name which should be the Republic of Chile in this case; in the attribute table, make sure Name is Republic of Chile, but everything else looks good!

When you are submitting multiple boundaries, you need to do multiple PRs and they each need to have their own branch (because the bot sometimes gets confused with more than one) so you may want to start 2 new PRs for these fixes and I'll close this PR to keep everything clean!

Issue #2786, opened 2023-03-15
[BOUNDARY ERRATA] HND ADM2 Nueva Frontera


mgt4:
The HND ADM2 open GeoBoundaries does not include the municipality of Neuva Frontera, in the department of Santa Barbara, Honduras. It is currently contained within the shape for Macuelizo municipality.

The total number of ADM2 administrative units (298) for Honduras appears correct however, because _Lago de Yojoa_ is included as a municipality. It is actually a lake and not considered an administrative unit of Honduras.

![image](https://user-images.githubusercontent.com/127971602/225366893-ed09eaa7-f665-4302-8a93-a7e72498be95.png)

PR #2825, opened 2023-04-11
hnd_adm2 update #2786

jingjli1:


amanda-reed (changes requested):
Hi Jing, just a few minor changes needed: for license source in your metadata, it should be the link where the license is specified which is the same as the link to source data in this case (it should match the link in the license screenshot). I also noticed a small separation between two parts of your boundaries, so unless that is intentional, you can put them together using the Align Features tool in arc (https://pro.arcgis.com/en/pro-app/latest/help/editing/align-features-to-a-traced-line.htm) Let me know if you have any questions!

amanda-reed:
image
Here's where it seems to be separated (and it continues down the entire boundary)

amanda-reed:
fixed and approved in #2828

Issue #1852, opened 2021-09-23
gB OPEN DOM_ADM1


lydiatroup:
need a new file for DOM ADM1 (no humdata, share-alike, or odbl)

PR #2836, opened 2023-04-21
DOM ADM1 #1852

jamesnelmore:
- Added NAM ADM1
- DOM ADM1 #1852


lydiatroup (changes requested):
Looks like you have both the NAM file and the DOM file here, can you re upload this with just the DOM file?

PR #2843, opened 2023-04-24
CAN_ADM3 #2025

amanda-reed:
 #2025 

submitted on behalf of @hadleyday


jingjli1 (changes requested):
just couple changes, column 'FEDNAME' should just be 'Name' and hide the shape length and shape area column if you can. @hadleyday

Sarangaban:
New file has been submitted with issues fixed

Issue #2771, opened 2023-02-26
gB OPEN LKA ADM0


lydiatroup:
Need a new file for LKA ADM0

jingjli1:
pr submitted and approved.

PR #2847, opened 2023-05-01
LKA_ADM0 #2771

emiliaz9:


amanda-reed (changes requested):
Hi Emilia, just a few little changes: 
- your metadata file needs to be named "meta" and the bot is case sensitive, so just be careful of that
- In your metadata, the License source should be the same as the source data in this case
-  your .atbx file is named IKA ADM0 for some reason, but I don't think we need this file, so you can just exclude it next time
- instead of including the source data, you need a screenshot of the license called "license" where the link is visible
Let me know if you have any questions!

jingjli1:
@emiliaz9  Please review the changes requested above and resubmit.

Sarangaban:
Changes made and new file submitted

PR #2939, opened 2023-08-14
Italy Adm0

rohith4444:


DanRunfola:
This doesn't appear to be an ADM0 - it has delienations.  ADM0 would be the country border.

PR #2940, opened 2023-08-14
Italy_adm1

rohith4444:


DanRunfola:
I am relatively certain this is the ADM2 for Italy.  It looks like all of these submissions are "downshifted" by one.

PR #3660, opened 2024-02-23
TZA_ADM1 gbOpen

geoBoundaryBot:
Boundary data for **TZA-ADM1** submitted through the geoBoundaries contribution form. 
    

**Name**: Emilia Zapata.
**Affiliation**: GeoBoundaries .
**Contact**: emiliaz0903@verizon.net.are files validated before they are uploaded to GitHub? Are there protections to ensure that the only expected file types/content can be uploaded?
**Notes about these data**:


Sarangaban:
Will speak to Emilia about it directly in the fall.

PR #3680, opened 2024-03-26
COG_ADM1 gbOpen

geoBoundaryBot:
Boundary data for **COG-ADM1** submitted through the geoBoundaries contribution form. 
    

**Name**: Julia Beverley.
**Affiliation**: GeoBoundaries @ William and Mary.
**Contact**: jbbeverley@wm.edu.are files validated before they are uploaded to GitHub? Are there protections to ensure that the only expected file types/content can be uploaded?
**Notes about these data**:


Sarangaban:
Few fixes. Will address in the fall - closed for now.

PR #3689, opened 2024-04-09
AUS_ADM0 gbOpen

geoBoundaryBot:
Boundary data for **AUS-ADM0** submitted through the geoBoundaries contribution form. 
    

**Name**: Julia Beverley.
**Affiliation**: GeoBoundaries @ William and Mary.
**Contact**: jbbeverley@wm.edu.are files validated before they are uploaded to GitHub? Are there protections to ensure that the only expected file types/content can be uploaded?
**Notes about these data**: This is actually for Norfolk Island, which is an external territory of Australia.


Sarangaban:
Will fix in the fall.

PR #3693, opened 2024-04-18
CRI_ADM3 gbOpen

geoBoundaryBot:
Boundary data for **CRI-ADM3** submitted through the geoBoundaries contribution form. 
    

**Name**: Jordan Lerner.
**Contact**: jblerner@wm.edu.are files validated before they are uploaded to GitHub? Are there protections to ensure that the only expected file types/content can be uploaded?
**Notes about these data**:


Sarangaban:
Cannot use license, will fix in the Fall.

PR #4139, opened 2024-09-23
ZMB_ADM2 gbOpen

geoBoundaryBot:
Boundary data for **ZMB-ADM2** submitted through the geoBoundaries contribution form. 
    

**Name**: Elizabeth Ott.
**Notes about these data**: I did not see the error where ADM 2 was listed in the attribute table twice, just to be sure I put "ADM2" in again for all but if that error pops up again you might have to show me what's going on so I can fix it.!


Sarangaban:
Closed. Resubmitted.

PR #4141, opened 2024-09-23
ARG_ADM1 gbOpen

geoBoundaryBot:
Boundary data for **ARG-ADM1** submitted through the geoBoundaries contribution form. 
    

**Name**: rohith.
**Affiliation**: geolab.
**Contact**: rmandala@wm.edu.are files validated before they are uploaded to GitHub? Are there protections to ensure that the only expected file types/content can be uploaded?
**Notes about these data**:


Sarangaban (commented):
Thanks Rohith! If you could add the ISO codes to the attribute table that would be great - thank you

rohith4444:
> Thanks Rohith! If you could add the ISO codes to the attribute table that would be great - thank you

Hi @Sarangaban, but i have added the ISO Codes for Entre Rios and Buenos Aires in attribute table before creating the pull request. I have attached screen shot as reference. Can you please cross-check and let me know.


Screenshot 2024-10-02 124727

PR #4187, opened 2024-12-02
VUT_ADM3 gbOpen

geoBoundaryBot:
Boundary data for **VUT-ADM3** submitted through the geoBoundaries contribution form. 
    

**Name**: Elizabeth Ott.
**Notes about these data**: The issue for this noted that the Name field was numbers rather than actual names. However, after further research it seems that Vanuatu's ADM3 boundaries do not have corresponding names, only numbers. Unsure if we want to keep this kind of boundary in the dataset, but I found an updated file so I'm updating it regardless.


Sarangaban (approved):
Great job, Thanks Elizabeth!

PR #4190, opened 2024-12-06
COL_ADM1 gbOpen

geoBoundaryBot:
Boundary data for **COL-ADM1** submitted through the geoBoundaries contribution form. 
    

**Name**: Elizabeth Ott.
**Notes about these data**: As of 2024, there is an area of dispute between the Huila and Cauca departments. This area was once part of the La Plata municipality, which is part of the Huila department. I merged the area of dispute back into the Huila province because the disputed area does not have a corresponding ISO code.


Sarangaban (approved):
Great job Elizabeth!

Issue #3868, opened 2024-05-01
gB Open SLE_ADM3 


Sarangaban:
Issue #2116 #2767

Make sure shape file has no holes within the boundary. Additionally resubmit meta data with geoBoundaries as source 1. Should be a quick fix.

ejott:
Resolved in #4192

PR #4191, opened 2024-12-06
SLE_ADM3 #3868

ejott:


ejott:
looks like i messed something up with metadata file, trying again

Issue #3919, opened 2024-05-03
gB Open LUX_ADM2


Sarangaban:
The canonical boundary name in meta data should be regions, county, or whatever the division is called instead of gbOpen, which is listed for release type. Another super quick fix.

PR #4193, opened 2024-12-06
LUX_ADM2 #3919

ejott:


ejott:
Second time today it's said there wasn't a meta.txt file included when there absolutely is one in the zipfile (submitting through Terminal), I will try to submit through website instead.

PR #4197, opened 2024-12-11
BWA_ADM1 gbOpen

geoBoundaryBot:
Boundary data for **BWA-ADM1** submitted through the geoBoundaries contribution form. 
    

**Name**: Sam Christophersen.
**Contact**: sjchristophers@wm.edu.are files validated before they are uploaded to GitHub? Are there protections to ensure that the only expected file types/content can be uploaded?
**Notes about these data**:


Sarangaban (changes requested):
Hi Sam looks great just need a few adjustments. Please add the canonical boundary name and the ISO codes. Thanks!

PR #4231, opened 2025-05-13
SLE_ADM2 gbOpen

geoBoundaryBot:
Boundary data for **SLE-ADM2** submitted through the geoBoundaries contribution form. 
    

**Name**: Elizabeth Ott.
**Contact**: ejott04@hotmail.com.are files validated before they are uploaded to GitHub? Are there protections to ensure that the only expected file types/content can be uploaded?
**Notes about these data**:


ejott:
Resolves part of issue #2116.

11helenab:
Need to update ISO codes, currently the attribute table has two columns with ADM level

